PR #113, opened 2013-10-03
fixes #109 - fire referenced controller init method

### LINKED ISSUE #109: BUG init() not called on pages referenced by SubsitesVirtualPage
When a SubsitesVirtualPage is added and viewed, the SubsitesVirtualPage_Controller init() method only fires the init() method of it's parent (VirtualPage_Controller). The init method of the referenced controller is not fired - so no initialisation code for the referenced page controller is fired either

moveforward:


tractorcow:
Closing in favour of fixing in core. See https://github.com/silverstripe/silverstripe-cms/issues/1386

PR #146, opened 2014-06-09
Changed the FileSubsites->canEdit method

laminbarrow:
It will now return true if the current user owns the file which would be very useful if you are allowing users to edit the file through and frontend form. The canEdit method will also check and return true if the current logged in user user has CMS_ACCESS_AssetAdmin or CMS_ACCESS_LeftAndMain permissions set


tractorcow:
Feedback ignored after 2 years so closing.

Issue #101, opened 2013-07-28
Creating a subsite doesn't update the subsite selection dropdown


hafriedlander:
From reporter:

In the CMS when a new subsite is created the CMS Admin cannot find it in the site menu (top left) it to add pages etc. It doesn't appear there until the user has logged out then back in again. 

This may leave the user confused as to why they cannot work on their new sub site. Note the guide makes no mention of this occuring

Issue #89, opened 2013-05-25
Hide global subsites dropdown when not applicable in specific CMS section


adrexia:
LeftAndMainSubsites.php Subsites method doesn't work as intended.

Because SilverStripe 3 doesn't rely on page reloads, menus won't change when navigating between sections. Given that menus are intended change, having a universal dropdown is a confusing ui element. Might be a better idea to have specific menus available in the sections of the site they are needed, and remove the global dropdown.

Issue #68, opened 2013-02-04
Subsite dropdown on a folder in AssetAdmin doesn't work


halkyon:
![Screen Shot 2013-02-04 at 3 39 10 PM](https://f.cloud.github.com/assets/138450/122604/4b800928-6e74-11e2-94ba-b15532743c0c.png)

This dropdown doesn't work. I assume it's supposed to switch the GridField below to the specific-subsite's files, but there's nothing to support this dropdown working at the moment.

PR #152, opened 2014-07-31
BUG: The move to subsite folder dropdown in files is gone

stojg:
I think the hiding of the dropdown and later re-enable from https://github.com/silverstripe/silverstripe-subsites/issues/68 from  #101 and #89 is broken and later the form name changed?


tractorcow:
Manually merged with https://github.com/silverstripe/silverstripe-subsites/commit/cef0781c926546e284aa7178b90c386aa1da25ae

PR #167, opened 2014-11-02
Automatically add main url to host-map.php

jonom:
If subsites is used in conjunction with [staticpublishqueue](https://github.com/silverstripe-labs/silverstripe-staticpublishqueue), _host-map.php_ currently has to be manually edited to include the main website's base URL every time a `Subsite` record is saved. Using these modules together already requires `FilesystemPublisher::$static_base_url` to be set, so we may as well make use of this setting if it is available to save this manual editing.


jonom:
Is there a way to re-run the tests?

stevie-mayhew:
@jonom - just repush with --force and they will go again.

jonom:
Thanks @stevie-mayhew - I had just tried that before but through a GUI so maybe it ignored me. I'll try from the command line

jonom:
Hmmm it's not working for me. Might have to send a new commit up

jonom:
Ah there we go. Had to do some silly business - added a bit of white space, committed, then removed the white space, committed again, then squashed the commits to get back to square one and force pushed.

... And it failed again.

Anyone know how to fix the test environment? Getting this message from Scrutinizer:

*We could not determine your dependencies. This is likely caused by a platform package (like a specific PHP version or a PHP extension) which is not available in the environment where PHP Analyzer runs. You can easily fix this by committing the composer.lock file to your repository.*

dhensby:
@tractorcow any advice on this?

tractorcow:
CORE_RELEASE=master is pointing to 4.0.x now, so it doesn't work with any module with ~3.1 dependency on framework.

Change CORE_RELEASE to 3 and it'll work on the 3 branch correctly.

tractorcow:
I've merged up some 1.0 changes into master and fixed the travis.yml... let's wait and see if travis likes these.

jonom:
@tractorcow Should I trigger a re-run of the tests again now?

tractorcow:
No it wouldn't help... the master branch of subsites hasn't been updated for a while, and since the 3.2 api has been re-"stabilised" it needs more update.

Did you want to use this feature with 3.1, by any chance?

tractorcow:
If someone wants to merge https://github.com/silverstripe/silverstripe-subsites/pull/181 and you rebase, then the tests will pass. :)

jonom:
It was intended for 3.1 but if it's easier to wait for 3.2 that's fine.

Do you know if there any active plans to update this module for 3.2 compatibility? It seems to have fallen a bit behind and is due for a bit of TLC (for instance it doesn't currently [support yml config](https://github.com/silverstripe/silverstripe-subsites/issues/174))

I can't currently run 3.2 on my local machine due to this issue: https://github.com/silverstripe/silverstripe-framework/issues/3426 so I can't help a lot with any changes targeting that version at the moment unfortunately.

tractorcow:
Hi @jonom you could try the MySQLPDODatabase if you are able to setup the PDO adaptor, since it doesn't run into the above issue.

This module should already be compatible with 3.2, but since the api isn't stable yet sometimes the module will need to be updated from time to time (as is the current case).

jonom:
Hmm still getting that 'We could not determine your dependencies' error.

tractorcow:
What does your composer.json look like?

jonom:
Mine? I just let the test run here so my composer.json shouldn't make a difference should it?

tractorcow:
Tests MIGHT pass if you rebase, @jonom . Want to try?

jonom:
I've rebased but tests are still failing - this PR must be cursed. I created a new SS site locally to test this patch and still seems to do the trick, site selection seems to work fine despite what behat test failure indicates. I'll try once more.

tractorcow:
Ouch, sorry. Will look at 4.x tests later and come back to this PR then. ;D

There were some changes to SS_Map in 4.x so those will need to be reflected here.

PR #169, opened 2014-11-11
Fix: ability to call actions on SubsitesVirtualPage

torleif:
The SubsitesVirtualPage_Controller disables actions to be called on the controller. This enables actions to be called as if it were the page in question.

### DIFF
--- a/code/SubsitesVirtualPage.php
+++ b/code/SubsitesVirtualPage.php
@@ -207,4 +207,13 @@ class SubsitesVirtualPage_Controller extends VirtualPage_Controller {
 		
 		Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
 	}
+	
+	public function allowedActions ($limitToClass = NULL) {
+		$parent = SiteTree::get()->byID($this->CopyContentFromID);
+		$model = $parent ? ModelAsController::controller_for($parent) : null;
+		if ($model && get_class($model) != get_class($this)) {
+			return array_merge($model->allowedActions(), parent::allowedActions ());
+		}
+		return parent::allowedActions ();
+	}
 }


code/SubsitesVirtualPage.php
```diff
@@ -207,4 +207,13 @@ public function init(){
 		
 		Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
 	}
+	
+	public function allowedActions ($limitToClass = NULL) {
```
dhensby: The argument to the function doesn't seem to be used

code/SubsitesVirtualPage.php
```diff
@@ -207,4 +207,13 @@ public function init(){
...
+		$parent = SiteTree::get()->byID($this->CopyContentFromID);
+		$model = $parent ? ModelAsController::controller_for($parent) : null;
+		if ($model && get_class($model) != get_class($this)) {
```
dhensby: If the classes match, then it must be another virtual page, yes?

If so, then shouldn't it take the actions from that virtual page (rather than just skipping as this code is doing).

code/SubsitesVirtualPage.php
```diff
@@ -207,4 +207,13 @@ public function init(){
...
+			return array_merge($model->allowedActions(), parent::allowedActions ());
```
dhensby: Spacing between parenthesis shouldn't be there

code/SubsitesVirtualPage.php
```diff
@@ -207,4 +207,13 @@ public function init(){
...
+		}
+		return parent::allowedActions ();
```
dhensby: Spacing between parenthesis shouldn't be there

code/SubsitesVirtualPage.php
tractorcow: It doesn't need to be, it's matching the method signature of the one it's overriding.

tractorcow:
No response in a year, so closing this.

Will address this issue with https://github.com/silverstripe/silverstripe-cms/issues/1386 in any case.

PR #172, opened 2014-11-24
redirect_fix_between_CMS_sections

pitchandtone:
previously if you were editing settings and you changed subsites ti would revert you to /admin, now it stays within your current controller


pitchandtone:
Is there any reason why this isn't merged?

dhensby:
Seems sensible, but I'm not familiar enough with subsites to take this in.

@tractorcow  ?

PR #176, opened 2014-12-16
Update SubsiteDomain.php

jason-zz:
Add $default_sort.

### DIFF
--- a/code/model/SubsiteDomain.php
+++ b/code/model/SubsiteDomain.php
@@ -8,6 +8,12 @@ class SubsiteDomain extends DataObject {
 
 	/**
 	 *
+	 * @var string
+	 */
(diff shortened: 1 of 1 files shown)


code/model/SubsiteDomain.php
```diff
@@ -8,6 +8,12 @@ class SubsiteDomain extends DataObject {
 
 	/**
 	 *
+	 * @var string
```
dhensby: please add `@config`

code/model/SubsiteDomain.php
```diff
@@ -8,6 +8,12 @@ class SubsiteDomain extends DataObject {
...
+	 */
+	private static $default_sort = "\"IsPrimary\" DESC";
```
dhensby: This should be single quotes so you don't need to escape the nested `"`

PR #177, opened 2015-02-01
ENHANCEMENT: option to hide global files

anselmdk:
Allowing to hide global folders in  `AssetAdmin`.

Config example:

    FileSubsites:
      show_global_folders: false

### DIFF
--- a/code/extensions/FileSubsites.php
+++ b/code/extensions/FileSubsites.php
@@ -9,6 +9,13 @@ class FileSubsites extends DataExtension {
 	// If this is set to true, all folders created will be default be
 	// considered 'global', unless set otherwise
 	static $default_root_folders_global = false;
+
+	/**
+	 * Show global folders in AssetAdmin
+	 * @config
+	 * @var bool
+	 */
+	private static $show_global_folders = true;
 	
 	private static $has_one=array(
 		'Subsite' => 'Subsite',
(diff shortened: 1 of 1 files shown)


dhensby:
I'm not entirely sure how closely this module wants to follow semver, but this is a new feature and so shouldn't make it into the 1.0 branch. Instead it should be against `master`.

anselmdk:
@dhensby thanks for reviewing this. Given that this is a non-breaking enhancement I believe it should be in the `1.0` branch. I wouldn't want to be forced to switch my `composer.json` to update from `master` for this, as _that_ would possibly introduce breaking changes to my codebase.
If you disagree with my reasoning, let me know, and I'll re-open a pull request to `master`.

dhensby:
For [SemVer](http://semver.org) a change of this kind has to go into the minor release (1.x) not a patch (1.0.x) and therefore this needs to be against master (which I assume is the 1.x release).

I can add a PR to alias master as 1.x so that you can do "~1.0" as a version constraint and get the master branch if your tracking dev dependencies.

dhensby:
Hmm, it does seem that master is tracking ss 3.2 and thus actually a 2.0.x release.

@tractorcow thoughts?

tractorcow:
I don't think we need to follow semantic versioning too closely, outside of the framework / cms (and their dependencies), although it's still a good model to follow. :)

My question is, how are global assets managed if they are permanently filtered out? Are admins not even allowed to view them, or are they just unrecoverable?

anselmdk:
@tractorcow the idea is that global files would be accessible from the main site, while site-specific files would be accessible from each subsite.
In the code `$subsiteID` would be 0 on the main site.

code/extensions/FileSubsites.php
```diff
@@ -9,6 +9,13 @@ class FileSubsites extends DataExtension {
 	// If this is set to true, all folders created will be default be
 	// considered 'global', unless set otherwise
 	static $default_root_folders_global = false;
+
+	/**
+	 * Show global folders in AssetAdmin
+	 * @config
+	 * @var bool
+	 */
+	private static $show_global_folders = true;
```
tractorcow: This config can't be turned off via YML, due to merge rules ignoring falsey values being merged with truthy.

Could we flip the meaning so that the default is false? E.g. 'hide_global_folders = false'.

tractorcow:
We are now following semver on this module, so um, can you please re-open this PR on master? Sorry to be a bummer!

anselmdk:
@tractorcow okay, I haven't worked on the affected project for a while. I'll revisit this, once I'll work on upgrading that project.

PR #179, opened 2015-02-11
redoing bug fix

pitchandtone:
60091c5b1714dc699c4f56d3a9da9d4e8b856212 got lost somehwere

### DIFF
--- a/code/SubsitesVirtualPage.php
+++ b/code/SubsitesVirtualPage.php
@@ -204,7 +204,15 @@ class SubsitesVirtualPage_Controller extends VirtualPage_Controller {
 		Subsite::$disable_subsite_filter = true;
 		
 		parent::init();
-		
+		// we also need to load the originally referenced page controller and fire it's init() method
+		$original_class = $this->CopyContentFrom()->ClassName;
+		// dont' run init() when page is newly created and does not have an original page assigned to it yet
+		if($original_class != 'SiteTree') {
+			$controller_class = $this->CopyContentFrom()->ClassName . '_Controller';
+			$cc = New $controller_class; // manually instantiate the class to stop 'method on non-object' errors
+			$cc->init();
+		}
+
 		Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
 	}
 }


pitchandtone:
Fixes #109

code/SubsitesVirtualPage.php
```diff
@@ -204,7 +204,15 @@ public function init(){
 		Subsite::$disable_subsite_filter = true;
 		
 		parent::init();
-		
+		// we also need to load the originally referenced page controller and fire it's init() method
+		$original_class = $this->CopyContentFrom()->ClassName;
+		// dont' run init() when page is newly created and does not have an original page assigned to it yet
+		if($original_class != 'SiteTree') {
+			$controller_class = $this->CopyContentFrom()->ClassName . '_Controller';
+			$cc = New $controller_class; // manually instantiate the class to stop 'method on non-object' errors
```
dhensby: Shouldn't this be `$cc = ModelAsController::controller_for($this->CopyContentFrom())`?

pitchandtone:
Yes it could be, I was just re-instating an old fix, but can update it.

dhensby:
please

dhensby:
In fact, there was a patch to 3.1 a while ago that means that a controller doesn't have to be defined any more for each page class so this has a very high likelihood of failing

tractorcow:
@pitchandtone please update so we can merge this. :)

dhensby:
maybe we should do this ourselves ;) @tractorcow

tractorcow:
I think @dhensby should do this ourselves.

dhensby:
:(

tractorcow:
I'm just wondering; Why is this fixed in the subsites module and not in VirtualPage_Controller core? It looks like the main controller has the same issue, and fixing it there would resolve the issue here as well.

tractorcow:
Oh wow, that's the 2175 special exception. I love that.

How about we move the ModelAsController setup from VirtualPage_Controller::__call (which is where it currently is) into VirtualPage_Controller, make sure it's init there.

We can replace the $this->__call('init') with $this->virtualisedController->init().

Instead of doing the 2175 hack, can't we do $this->hasMethod() it VirtualPage_Controller, and then fall back to $this->virtualisedController->$method($args);

Why instead of everything can't virtual pages just be easy? I guess not...

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/commit/1eb80b7498e91362f34212eb8ca2b0c51c14f628

dhensby:
classic, so what's the plan, then?

tractorcow:
Maybe I'll look at it on hackday. It's ugly enough that it would make me feel better to fix it.

tractorcow:
I'm going to close this and re-open on cms module.

tractorcow:
Moved to https://github.com/silverstripe/silverstripe-cms/issues/1386

PR #183, opened 2015-03-23
Fix subsites to use correct permissions

tractorcow:
See http://www.silverstripe.org/software/download/security-releases/ss-2015-008-sitetree-creation-permission-vulnerability

Logged out users can never create pages (fixed in cms)

CMSMain controls pages creatable at the root using SiteTreeHints, as the PageTypes no longer assumes knowledge of the ability to create pages.

### DIFF
--- a/tests/SiteTreeSubsitesTest.php
+++ b/tests/SiteTreeSubsitesTest.php
@@ -8,6 +8,10 @@ class SiteTreeSubsitesTest extends BaseSubsiteTest {
 		'SiteTreeSubsitesTest_ClassA',
 		'SiteTreeSubsitesTest_ClassB'
 	);
+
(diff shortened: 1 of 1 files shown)


tests/SiteTreeSubsitesTest.php
```diff
@@ -8,6 +8,10 @@ class SiteTreeSubsitesTest extends BaseSubsiteTest {
 		'SiteTreeSubsitesTest_ClassA',
 		'SiteTreeSubsitesTest_ClassB'
 	);
+
+	protected $illegalExtensions = array(
```
tractorcow: This was added to simplify testing of this module with translatable included.

Possibly we could add subsites + translatable specific tests in the future.

PR #194, opened 2015-07-15
Pass $action through the alternateAbsoluteLink() method

threesquared:
This allows you to use `->AbsoluteLink('action')`

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -211,10 +211,10 @@ class SiteTreeSubsites extends DataExtension {
 		}
 	}
 
-	function alternateAbsoluteLink() {
+	function alternateAbsoluteLink($action = null) {
 		// Generate the existing absolute URL and replace the domain with the subsite domain.
 		// This helps deal with Link() returning an absolute URL.
-		$url = Director::absoluteURL($this->owner->Link());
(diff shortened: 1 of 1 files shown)


code/extensions/SiteTreeSubsites.php
```diff
@@ -208,10 +208,10 @@ static function contentcontrollerInit($controller) {
 		if($subsite && $subsite->Theme) SSViewer::set_theme(Subsite::currentSubsite()->Theme);
 	}
 
-	function alternateAbsoluteLink() {
+	function alternateAbsoluteLink($action) {
```
dhensby: this needs to be an optional parameter, I'd have thought?

dhensby:
I think this would be useful against 3 / 3.2 and not just master.

code/extensions/SiteTreeSubsites.php
threesquared: An argument is passed through from the call in SiteTree (https://github.com/silverstripe/silverstripe-cms/blob/3.2/code/model/SiteTree.php#L477) but you are right, better to be safe.

tractorcow:
Please rebase on master.

wernerkrauss:
I think this be closed, as it's fixed in v4 now. Thanks for your effort @threesquared

PR #196, opened 2015-08-01
made duplication between subsites more robust

pitchandtone:
and added some extra methods so it's easier to overwrite or extend

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -169,35 +169,80 @@ class SiteTreeSubsites extends DataExtension {
 	}
 
 	/**
+	 * Does the basic duplication, but doesn't write anything
+	 * this means we can subclass this easier and do more complex
+	 * relation duplication.
+	 */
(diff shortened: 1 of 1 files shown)


pitchandtone:
Unsure why scrutinizer is erroring. Anyone?

tractorcow:
We need to fix the tests to not run behat from master.

tractorcow:
Please rebase on master.

dhensby:
ping

tractorcow:
pong!

pitchandtone:
Rebased.

dhensby:
@tractorcow master tests are failing anyway, but these failures seem worse?

This change looks significant to require a test, I think

tractorcow:
Someone needs to fix these tests... master is too fragile at the moment. =(

pitchandtone:
@dhensby @tractorcow do you want me to look at these tests? Or have you guys already?

tractorcow:
We'll fix the tests up later. We accept that modules running against master of framework won't pass for a while, so no need to let them block merges.

PR #204, opened 2015-10-28
Added missing route to `SubsiteXHRController`

### LINKED ISSUE #203: CMS requests to SubsiteXHRController fail with 404.
I'm using `1.1.x-dev` and SilverStripe 3.2.
Navigating the CMS triggers requests to `SubsiteXHRController` which all fail with a 404.

As much as I can tell, these are triggered via JavaScript from `LeftAndMain_Subsites.js`. I think there's a route that's missing in a config file?

Adding something like this makes the problem go away:
```
Director:
  rules:
    'SubsiteXHRController': 'SubsiteXHRController'
```

Not sure if that wasn't required prior to 3.2, because on a first glance I wasn't able to find the route in previous versions?

bummzack:
Ensure SilverStripe 3.2 compatibility. Fixes #203

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -4,3 +4,7 @@ After: 'framework/*','cms/*'
 ---
 AssetAdmin:
   treats_subsite_0_as_global: true
+
+Director:
+  rules:
+    'SubsiteXHRController': 'SubsiteXHRController'


dhensby:
hmm, I'm surprised we don't have a failing test... let me look into this

bummzack:
@dhensby Because test still use auto-routing. From the 3.2 changelog: *The auto-routing is still in place for unit tests, since its a frequently used feature there*

dhensby:
Classic!

dhensby:
thanks :)

PR #221, opened 2015-11-24
API Add option to specify http / https on subsite domains

tractorcow:
Built on top of https://github.com/silverstripe/silverstripe-subsites/pull/220 because I expected it would have been merged around the same time. :D

This is what the UX now looks like:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 
 language: php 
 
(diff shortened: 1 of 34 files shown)


tractorcow:
On advice I have renamed "both" to "automatic"

PR #222, opened 2015-11-26
BUG Fix menu compatibility with framework 4.x

tractorcow:
Synchronises this module with the upstream framework template.

I've also created https://github.com/silverstripe/silverstripe-framework/pull/4795 to help minimise the impact of these kinds of issues in future, but neither depends on the other. (you can merge one or the other safely without regression).


tractorcow:
Ta. :D

PR #224, opened 2015-12-20
Moved user docs into userguide folder to display on userhelp site

mandrew:
moved user content into userhelp folder, removed old user manual pdf, added index file, updated readme links, moved images.

**3.2 compatible docs**


tractorcow:
This likely won't get merged up to 1.2 or master for a while. I'd create it on master if I were you...

mandrew:
@tractorcow this will still need to be merged into 1.1 though... also what is the current stable branch for subsites? Is it 1.2.0? We were updating the documentation on the basis that 1.1 was the stable.

If 1.1 is the current stable then I need a confirmation that these changes will be merged into the 1.2 when it becomes stable?

If 1.2 is the current stable then these changes will also need to be merged into 1.2 and once done I will need to update the link from the docs repo

tractorcow:
It is 1.2, yes. :)

Feel free to cross-pull request this change to each branch if you want it merged earlier.

PR #225, opened 2015-12-20
Moved user docs into userguide folder to display on userhelp site

mandrew:
moved user content into userhelp folder, removed old user manual pdf, added index file, updated readme links, moved images.

** 3.1 compatible docs **


tractorcow:
Are you going to publish multiple versions of the user docs for this module? Otherwise it doesn't seem to add much value.

mandrew:
@tractorcow Already done and live, just pointing to my repo so I want to merge my changes in and point the code to the module. There are different versions of the documentation https://userhelp.silverstripe.org/en/3.2/optional_features/working_with_multiple_sites/ and https://userhelp.silverstripe.org/en/3.1/optional_features/working_with_multiple_sites/.

I don't think there are huge differences between each but it allows for future changes to the module without editing previous versions

tractorcow:
Well, ok :)

Just so you know, 1.0 is the lastest 3.1 compatible branch, but 1.2 is the latset 3.2 compatible branch. Might want to check your branches for 3.2.

PR #237, opened 2016-02-10
Fix error when dataQuery is a null object

mparkhill:
Add additional check on dataQuery to prevent the error message: 'attempting to call a function on a non-object' in certain cases where $dataQuery variable is null.

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -33,7 +33,8 @@ class SiteTreeSubsites extends DataExtension
         if (Subsite::$disable_subsite_filter) {
             return;
         }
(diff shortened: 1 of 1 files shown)


tractorcow:
I think if $dataQuery is null, there is another problem still to solve.

Have you done a backtrace to confirm where this is occurring?

DataQuery.php has `$obj->extend('augmentSQL', $query, $this);` and DataObject.php has

```php
$query = $dataQuery->query();
$this->extend('augmentLoadLazyFields', $query, $dataQuery, $this);
$this->extend('augmentSQL', $query, $dataQuery);
```

Both of which couldn't possible have a null value for $dataQuery.

dhensby:
Bump

Issue #235, opened 2016-02-05
Subsite Virtual Page won't use subsites template


torleif:
Recently the subsite module stopped using templates specified by the Subsite Virtual Page. For example with the following site structure: 

themes/maintheme/Layout/MyPageType.ss
themes/mysubsite/Layout/MyPageType.ss

A subsiteVirtualPage with the linked page of MyPageType.php will be rendered with the Page.ss of mysubsite, instead of MyPageType.ss

tractorcow:
Added to the big bag of "virtual page issues to fix"

dhensby:
@torleif have you PR-ed your commit?

torleif:
@dhensby PR-ed?

dhensby:
Pull request. Did you open a pill request with the commit that references this issue?

torleif:
@dhensby nope, i'll do that now

robbieaverill:
Do we need an epic type holder issue for virtual page bugs?

robbieaverill:
Assuming this still affects v4, I'm retagging

robbieaverill:
I can't reproduce this on SilverStripe 4.4.x with Subsites 2.3.x. I used the CWP starter and watea themes, where my subsite has watea configured and starter contains silverstripe/blog page templates where watea doesn't. The virtual page correctly uses the starter blog page templates, rather than reverting to a Page template (which watea _does_ contain).

Unfortunately, [SilverStripe 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for SilverStripe 3 going forward.

You can read the [SilverStripe Roadmap](https://www.silverstripe.org/software/roadmap/) for more information on our support commitments.

PR #241, opened 2016-03-22
Fix: SubsiteVirtualPage will use subsite template #235

torleif:
This is a potential fix for #235

### DIFF
--- a/code/SubsitesVirtualPage.php
+++ b/code/SubsitesVirtualPage.php
@@ -230,4 +230,12 @@ class SubsitesVirtualPage_Controller extends VirtualPage_Controller
         
         Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
     }
+    
+    public function getViewer($action) {
+        $originalClass = get_class($this->CopyContentFrom());
+        if ($originalClass == 'SiteTree') $name = 'Page_Controller';
+        else $name = $originalClass."_Controller";
+        $controller = new $name($this->CopyContentFrom());
+        return $controller->getViewer($action);
+    }
 }


code/SubsitesVirtualPage.php
```diff
@@ -230,4 +230,12 @@ public function init()
         
         Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
     }
+    
+    public function getViewer($action) {
+        $originalClass = get_class($this->CopyContentFrom());
+        if ($originalClass == 'SiteTree') $name = 'Page_Controller';
+        else $name = $originalClass."_Controller";
+        $controller = new $name($this->CopyContentFrom());
```
dhensby: This isn't the correct way to get a controller for a model.

use `ModelAsController`

```php
$controller = ModelAsController::controller_for($this->CopyContentFrom());
$controller->init();
return $controller->getViewer($action);
```

code/SubsitesVirtualPage.php
```diff
@@ -230,4 +230,12 @@ public function init()
         
         Subsite::$disable_subsite_filter = $origDisableSubsiteFilter;
     }
+    
+    public function getViewer($action) {
+        $originalClass = get_class($this->CopyContentFrom());
+        if ($originalClass == 'SiteTree') $name = 'Page_Controller';
```
tractorcow: Can you please use braces for if-else blocks.

code/SubsitesVirtualPage.php
tractorcow: Or just do this. :)

PR #246, opened 2016-05-16
Set sudo:false to use new containerised structure

tractorcow:
I can't see much reason for 1.2 behat tests passing but 1.1 failing, so seeing if this fixes it.


tractorcow:
Wow, magic. :D

Issue #53, opened 2012-08-24
Unique urls enforced across sites


tazzydemon:
Have been swapping between the tractorcow repo and this plus I am using the new Multilingual plugin from kreationsbyran.se. I am therefore in at the deep end so to speak 

However I have noticed that the subsites module does not override the built-in requirement to have unique urls. Therefore across domains it is the same, i.e

domain1.com/home
domain2.com/home is forced to be domain2.com.home-2

IMHO This is not the behaviour that would be expected.

tractorcow:
This doesn't fix the issue properly, so I've re-opened.

This only fixes the issue where the page being saved is in the current subsite. If the page belongs to another subsite, then it will (mistakenly) be filtered in the current subsite.

The correct fix is to implement augmentValidURLSegment() and to check it manually.

In framework 4.x we have the concept of "inheritable" query parameters; We can probably update the root SiteTree::validateURLSegment method so that it inherits these parameters (including subsite) from the parent object being checked.

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-subsites/pull/247

PR #247, opened 2016-05-23
BUG Fix issue with urlsegment being renamed in subsites

### LINKED ISSUE #193: Subsite Virtual Page URLSegment Iterator Issue
Making changes to a page triggers the duplicate URLSegment itarator to trigger on any VirtualPages pointing to it.

For istance if I change a page with the URL Segment my-page all it's subsite virtual pages become my-page-2

tractorcow:
With this fix in place, validateURLSegment() will no longer check for conflicts against pages that belong to other subsites.

I've also removed the custom logic from SubsiteVirtualPage: It didn't work properly, and the substituted logic should work better. :) This is covered by existing tests.

Fixes #193 and #53 

The code in this module is super-due for a rewrite soon.

### DIFF
--- a/code/SubsitesVirtualPage.php
+++ b/code/SubsitesVirtualPage.php
@@ -9,18 +9,18 @@ class SubsitesVirtualPage extends VirtualPage
         'CustomMetaDescription' => 'Text',
         'CustomExtraMeta' => 'HTMLText'
     );
-    
+
     public function getCMSFields()
     {
         $fields = parent::getCMSFields();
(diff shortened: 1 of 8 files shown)


tractorcow:
Also, this bug is 8 years old? http://www.silverstripe.org/community/forums/archive/show/3031

dhensby:
Let's get this into the wild!

tractorcow:
This module is already pretty wild.

PR #250, opened 2016-05-27
BUG Fix copy to subsite breaking on sub-pages

### LINKED ISSUE #192: Nested pages copied to subsite are inaccessible
When copying a nested page (i.e one that isn't at the root level) to a subsite, the copied page is created however it appears to try and keep the original parent ID which means it is not accessible as the parent is not in the same site.

tractorcow:
Fixes #192

jira ticket at https://silverstripe.atlassian.net/browse/CWP-448

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -84,13 +84,17 @@ class SiteTreeSubsites extends DataExtension
         if ($isDefaultSubsite && $subsitesMap) {
             $fields->addFieldToTab(
                 'Root.Main',
-                new DropdownField(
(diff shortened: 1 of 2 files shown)


code/extensions/SiteTreeSubsites.php
```diff
@@ -208,22 +212,21 @@ public function canPublish($member = null)
 
     /**
      * Create a duplicate of this page and save it to another subsite
-     * @param $subsiteID int|Subsite The Subsite to copy to, or its ID
+	 *
```
chillu: Wrong indentation, using tabs rather than spaces (looks like we've converted the module to PSR-2 already?)

PR #256, opened 2016-07-18
Fix error when dataQuery is a null object

madmatt:
Credit to @mparkhill who fixed this originally, but never raised a pull request to get the fix back to the base module.

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -33,7 +33,7 @@ class SiteTreeSubsites extends DataExtension
         if (Subsite::$disable_subsite_filter) {
             return;
         }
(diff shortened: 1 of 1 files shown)


tractorcow:
When does augmentSQL ever not get passed a second parameter?

tractorcow:
I asked this at https://github.com/silverstripe/silverstripe-subsites/pull/237#issuecomment-182692758 but never got an answer.

dhensby:
@tractorcow that's a good point - @madmatt do you know what circumstances this happens?

Personally, I'm happy with this kind of "defensive" coding regardless, but feel free to revert if you like, @tractorcow

tractorcow:
Don't worry, the quality of subsites isn't degraded at all by this change. :) Not on average, haha.

PR #264, opened 2016-09-22
SS4 namespaces compatibility

kinglozzer:
There’s still a lot of work needs to be done to make this work properly, but at least a `dev/build` will run now. I’ve run the upgrade tool on it and fixed a few things it misses (YAML, template paths etc)


dhensby:
`Class 'SilverStripe\AssetAdmin\Controller\AssetAdmin'`

Seems like a relevant error? `AssetAdmin` is no longer a thing?

sminnee:
> Seems like a relevant error? AssetAdmin is no longer a thing?

This is a sign we need to install the asset-admin module as part of this build. Best fix is probably to make it a requirement of the subsites module?

kinglozzer:
Yep, that should cover it. I’ve added that now, and removed the controller that I broke in https://github.com/silverstripe/silverstripe-cms/pull/1605. Tests are still failing, I’ll look into it - I’m still wrestling with why some CMS bits aren’t working (changed templates, classes etc)

dhensby:
@kinglozzer we don't run the CMS tests against framework any more (or, at least, they are allowed failures) so failures may be expected?

kinglozzer:
I’ve fixed a few more bits of functionality (haven’t looked at the tests), but I’m going to be short on time over the coming week or so. This could be merged now – it’s better than the current state, which is completely broken – or if anyone else fancies adding to this PR I’ve enabled “edits from maintainers”.

kinglozzer:
Stuff that’s been done:
- Added import statements/converted string class names (there may be one or two that still need to be updated in tests)
- Fixed the template path for `LeftAndMain_Menu`
- Fixed a few issues with `updateCMSFields()` in the extensions

Stuff that still needs to be done:
- Check/fix the other template paths
- Fix tests
- Namespace the subsites classes themselves (not _required_, but should be done ideally)
- Move extension registration from `_config.php` to `config.yml` (as above, this is a nice-to-have)

dhensby:
TBH - there's a whole lot of work that can be done to subsites to make it nicer ;)

Issue #261, opened 2016-09-12
Option to run multiple sites / subsites without theme (theme-less)


johanchristian:
Selecting a theme is currently mandatory, when adding a new subsite.
Can we please have the option to make this optional?

Thank you.

robbieaverill:
## Scenarios

### No themes installed

As a CMS user,
Given that I have no themes installed,
And I add a subsite,
I expect not to see a Theme dropdown,
So my user interface is not unnecessarily cluttered

### One or more themes installed

As a CMS user,
Given that I have one or more themes installed,
And I add a subsite,
I expect to see a Theme dropdown,
And I expect it to have an empty option,
So that I can either choose a theme or not

robbieaverill:
Fixed in #273

PR #265, opened 2016-11-04
Add setEmptyString to prevent error when no themes

RuthAdele:
To fix issue #261


dhensby:
I think it's confusing for a CMS user to see a dropdown with "please select" and no options. Can we wrap this in a `if ($this->allowedThemes)` ?

sminnee:
Yeah I think that if there are no themes, the whole dropdown should be removed from the form.

robbieaverill:
@dhensby @sminnee we have a fix that covers those changes at #273 - do you think it qualifies as a patch update so can safely go into the 1.3 branch?

sminnee:
@robbieaverill it's a new feature so probably needs to go into the `1` branch.

sminnee:
Replaced with #273

PR #266, opened 2016-11-11
Fix copying child pages to subsite

flashbackzoo:
Copying child pages to a subsite is currently broken.

screen shot 2016-11-11 at 4 23 15 pm

Currently the parent page is copied but not the children. This is because `$this->owner->AllChildren()` is called _before_ `Subsite::changeSubsite($oldSubsite)` which means the children of the _copied_ Page are being returned from `AllChildren()`, and there are none.

Moving `Subsite::changeSubsite($oldSubsite)` up fixes that issue. But then all child pages are copied to the root level. So I've added a param for the parent ID.


dhensby:
Seems sensible to me. I'm happy this can go into 1.3.x

PR #269, opened 2017-01-25
Update PHPUnit version to 4.8

robbieaverill:
As in the framework


dhensby:
lots of test failures :/ are they pre-existing?

robbieaverill:
@dhensby it looks like it against the master branch, yeah

PR #270, opened 2017-02-03
CMS fix for Group Extension

gaethofs:
ArrayList()->map() returns SilverStripe\ORM\Map.
Needs to be converted to an array.


dhensby:
Looks like master needs a lot of love on the test front :/

tractorcow:
Superceded with @robbieaverill 's work that's been recently merged.

PR #271, opened 2017-02-10
Fix subsites hiding menu items in the cms.

ajoneil:
alternativeMenuDisplayCheck() should return true in order for
it to flow through to the normal permissions checks (see
https://github.com/silverstripe/silverstripe-framework/blob/3.5/admin/code/LeftAndMain.php#L650)


ajoneil:
Hmm, this looks like a case of "read the docs", looks like cms sections are deliberately disabled by default unless you opt-in.

PR #273, opened 2017-04-10
Conditionally add Theme dropdown (if there are any themese available)…

danaenz:
… for #261.

PR #265 grew a little stale, so reopening a new PR to move this along

### DIFF
--- a/code/model/Subsite.php
+++ b/code/model/Subsite.php
@@ -583,15 +583,15 @@ class Subsite extends DataObject
      */
     public function getCMSFields()
     {
-        if ($this->ID!=0) {
-            $domainTable = new GridField(
+        if ($this->ID != 0) {
+            $domainTable = GridField::create(
                 "Domains",
                 _t('Subsite.DomainsListTitle', "Domains"),
                 $this->Domains(),
                 GridFieldConfig_RecordEditor::create(10)
             );
         } else {
-            $domainTable = new LiteralField(
+            $domainTable = LiteralField::create(
                 'Domains',
                 '<p>'._t('Subsite.DOMAINSAVEFIRST', 'You can only add domains after saving for the first time').'</p>'
             );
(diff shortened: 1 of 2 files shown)


code/model/Subsite.php
```diff
@@ -648,7 +645,12 @@ public function getCMSFields()
             new HiddenField('IsSubsite', '', 1)
         );
 
-        $subsiteTabs->addExtraClass('subsite-model');
+		// If there are any themes available, add the dropdown
+		if (!empty($this->allowedThemes())) {
```
robbieaverill: We need to support PHP 5.3/5.4 in this branch, so will need to change to:

```
$allowedThemes = $this->allowedThemes;
if (!empty($allowedThemes)) {
    // ... 
```

code/model/Subsite.php
```diff
@@ -648,7 +645,12 @@ public function getCMSFields()
...
+			$fields->addFieldToTab('Root.Configuration', new DropdownField('Theme', $this->fieldLabel('Theme'), $this->allowedThemes(), $this->Theme), 'PageTypeBlacklistToggle');
```
robbieaverill: I know this is pre-existing, but could you swap for `DropdownField::create` so it's injectable?

danaenz:
@robbieaverill I've converted the existing Field constructors in the getCMSFields() function to use injection, however the $this->allowedThemes is not compatible with PHP 5.6 - it returns null ($this->allowedThemes() returns the array). $this->allowedThemes() is calling a public function, so I'm not sure why it would be incompatible with 5.3/5.4? Would you be able to provide some clarification here so I can test and make updates?

robbieaverill:
@danaenz if you take a look at the Travis failures on PHP 5.3 and 5.4 you'll see why this failed (https://travis-ci.org/silverstripe/silverstripe-subsites/jobs/220405623) - I think you should stick to using the public method rather than the property (personally)

code/model/Subsite.php
```diff
@@ -610,45 +610,48 @@ public function getCMSFields()
...
-            new HiddenField('IsSubsite', '', 1)
+            HiddenField::create('ID', '', $this->ID),
+            HiddenField::create('IsSubsite', '', 1)
         );
 
-        $subsiteTabs->addExtraClass('subsite-model');
+		// If there are any themes available, add the dropdown
+		$themes = $this->allowedThemes();
+		if (!empty($themes)) {
+			$fields->addFieldToTab('Root.Configuration', DropdownField::create('Theme', $this->fieldLabel('Theme'), $this->allowedThemes(), $this->Theme), 'PageTypeBlacklistToggle');
```
robbieaverill: Minor: may as well re-use `$themes` instead of `$this->allowedThemes()` here. Could you squash the commits? Looks good :)

danaenz:
Commits squashed. :)

robbieaverill:
Sorry for the delay @danaenz. I've added a couple of user stories to #261 which I think covers the scenarios I can see.

* This PR is good, but would you mind adding an empty string (literally) so people don't have to choose a theme if they don't want to?

The areas of the code using the theme property are adequately checking for a set value already, so the outcome from #261 should be covered by this PR with the empty string added in (originally #265) and added to this PR too.

danaenz:
Hi @robbieaverill , I've added the empty string plus a lang entry for that. Squashed commits as well

code/model/Subsite.php
```diff
@@ -610,45 +610,49 @@ public function getCMSFields()
...
+            HiddenField::create('ID', '', $this->ID),
+            HiddenField::create('IsSubsite', '', 1)
         );
 
-        $subsiteTabs->addExtraClass('subsite-model');
+		// If there are any themes available, add the dropdown
+		$themes = $this->allowedThemes();
+		if (!empty($themes)) {
+			$fields->addFieldToTab('Root.Configuration',
+				DropdownField::create('Theme', $this->fieldLabel('Theme'), $this->allowedThemes(), $this->Theme)->setEmptyString(_t('Subsite.ThemeFieldEmptyString', 'Default')), 'PageTypeBlacklistToggle');
```
robbieaverill: Sorry, I meant a literal empty string. Translation would be useful though if people want to set something like "Default"

danaenz: Could you clarify? Do you mean, pass an empty array as the source?

setEmptyString uses an empty value, and this has a translatable label, so it seems like an adequate fix to me.

I have noticed that a check is failing so I'll have a look into that too.

robbieaverill: @danaenz I'd expect to see an empty option in the dropdown field, then an option for each available theme. I think the request in the original issue was to not have to select a theme, which an empty option/string implies, whereas "Default" _does_ imply the use of a theme - being the default theme - even though it'll be saved the same way in the database.

danaenz:
I've set it to have an empty label as well - the first option in the dropdown will be blank, and will then be followed by the available themes.

lang/en.yml
```diff
@@ -72,6 +72,7 @@ en:
     PrimaryDomainFieldLabel: 'Primary Domain'
     RedirectURLFieldLabel: 'Redirect URL'
     ThemeFieldLabel: Theme
+    ThemeFieldEmptyString: 'Default'
```
robbieaverill: This would take precendence - can you set this to an empty string too?

danaenz:
Oops - forgot about that one! Should be there now. :)

robbieaverill:
Last thing @danaenz can you switch the branch in this PR to `1` and rebase if necessary?

sminnee:
I've created a 1 branch and moved this to it. @robbieaverill over to you

robbieaverill:
@danaenz the test failure seems related (has occurred over multiple builds on PHP 5.6) https://travis-ci.org/silverstripe/silverstripe-subsites/jobs/225060640

danaenz:
@robbieaverill I'm having trouble debugging this particular one. The tests pass on my environment (PHP 5.6.9). My PHPUnit version is later, ~5.7.

I'll keep trying!

robbieaverill:
@danaenz looks like the build failure was temporary (that build line runs against the 3.x dev branch of framework). I've re-run it and it's all green now. Thanks!

PR #277, opened 2017-05-22
Update SQL to explicitly use link table column

DarrenInwood:
Explicitly use tablename in SQL to avoid column ambiguity.

Avoids errors when Group or Member has had a MemberID column added via extension.


DarrenInwood:
Forked incorrectly, trying again

PR #288, opened 2017-08-10
Fix page rendering with proper subsite locale

### LINKED ISSUE #83: subsite ignores language
in ss 3.1, configure two subsites, one in english, one in german. your site's default lang is english.

Despite having configured the german subsite to use "German" as language, the created HTML (the simple theme uses the field $ContentLocale) is set to english (en-US)

gregsmirnov:
Fixes #83


robbieaverill:
Thanks @gregsmirnov!

It should be noted for users of CWP that Translatable will take priority in deciding the ContentLocale over this patch. In this case they'd need to ensure that they've also set the correct language in Translatable for their German subsite.

This patch fixes the problem for sites not using Translatable.

PR #291, opened 2017-08-29
Update Travis configuration for PHP 5.6-7.1, add composer dependencies and codecov.io config

robbieaverill:
Note: contains hacks to make the tests run:

* Exclude behat test dir from manifest
* Add simple namespace imports to top level tests in `tests/` folder
* PSR-2 linting
* Test class API changes implemented

PR #292, opened 2017-08-29
FIX Run SS standard PHP linter, separate SiteTree test mock classes

robbieaverill:
Also updates some test class API changes and moves classes in the same file into separate files


robbieaverill:
Closed, merged into #291

PR #293, opened 2017-08-29
NEW Add SubsiteState and initialisation middleware

robbieaverill:
This adds a SubsiteState service which contains the subsite ID and can be accessed globally, adds detection via a middleware and replaces `Subsite::currentSubsiteID` use with it.

Added deprecation version and notice to `Subsite::currentSubsiteID`.

### DIFF
--- a/_config/middleware.yml
+++ b/_config/middleware.yml
@@ -1,0 +1,10 @@
+---
+Name: subsitesmiddleware
+After:
+  - requestprocessors
+---
+SilverStripe\Core\Injector\Injector:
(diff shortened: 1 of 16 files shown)


_config/middleware.yml
```diff
@@ -0,0 +1,8 @@
+---
+Name: subsitesmiddleware
```
tractorcow: Can you make sure this is initialised after the session middleware?

robbieaverill: Updated, thanks

PR #294, opened 2017-08-31
Fix up test suite, re-enable Behat, update further SS4 API changes

robbieaverill:
I've marked the SubsitesVirtualPage tests as incomplete as they're currently failing, and my time box  ran out 😄  will log an issue to potentially remove it if it's not used any longer, or otherwise to update it from CMS and fix up the tests.


code/State/SubsiteState.php
```diff
@@ -37,6 +43,36 @@ public function setSubsiteId($id)
     {
         $this->subsiteId = (int) $id;
 
+        // Persist to session, if they are enabled
+        if ($this->getUseSessions() && Injector::inst()->has(HTTPRequest::class)) {
+            Injector::inst()
+                ->get(HTTPRequest::class)
+                ->getSession()
+                ->set('SubsiteID', $id);
+        }
```
robbieaverill: There is a problem with this in that the session is not nested with using `->withState()` callbacks, so by setting the subsite ID in this way it would remain in the session afterwards.

Could we potentially nest the Injector in withState to avoid this?

tractorcow: You should put this in the middleware then

eg..

```
class Middleware {
	function process($request, $next) {
		$state = SubsiteState::singleton();
   		$state->setSubsite($request->getSession()->get('SubsiteID'));
		try {
			return $next();
		} finally {
			$request->getSession()->set('SubsiteID', $state->getSubsiteID());
		}
}
```

That way 100% of your session access is in middleware.

robbieaverill: Cool, updated

robbieaverill:
Noting issues I've found so far during smoke testing:

* Asset admin doesn't work the same way as in SS3, so we'll need to come up with a way of allowing the FileSubsites extension to either filter the file list, visually indicate that a file is not visible in a given subsite at that point, or add a dropdown to the details screen to switch subsite

code/Middleware/InitStateMiddleware.php
```diff
@@ -2,46 +2,96 @@
...
+        $state->setUseSessions($isAdmin);
+
+        // Detect the subsite ID
+        $subsiteId = $this->detectSubsiteId($request);
+        $state->setSubsiteId($subsiteId);
+
+        // Persist to the session if using the CMS
+        if ($state->getUseSessions()) {
+            $original = $request->getSession()->get('SubsiteID');
+            $request->getSession()->set('SubsiteID', $subsiteId);
```
tractorcow: You should be calling this `Session::set` after the `$delegate($request)`, not immediately after getting the value from the session.

code/extensions/ControllerSubsites.php
```diff
@@ -15,7 +15,7 @@ public function controllerAugmentInit()
     {
         if ($subsite = Subsite::currentSubsite()) {
             if ($theme = $subsite->Theme) {
-                SSViewer::set_themes([$theme, SSViewer::DEFAULT_THEME]);
+                SSViewer::add_themes([$theme]);
```
tractorcow: You don't want $theme to be after the default; Otherwise your theme may not even get activated for many templates!

code/Middleware/InitStateMiddleware.php
```diff
@@ -27,28 +28,35 @@ class InitStateMiddleware implements HTTPMiddleware
...
-            // Track session changes
-            $state->setSessionWasChanged($subsiteId === $original);
-        }
+            return $delegate($request);
+        } catch (DatabaseException $ex) {
+            // No-op, database is not ready
+        } finally {
+            if ($state->getUseSessions()) {
+                // Track session changes
+                $request->getSession()->set('SubsiteID', $subsiteId);
```
tractorcow: If subsite id is changed in $delegate, you'll need to get the current subsite back from the state, not the value of $subsiteID which will be out of date.

code/Middleware/InitStateMiddleware.php
```diff
@@ -27,28 +28,35 @@ class InitStateMiddleware implements HTTPMiddleware
...
 
-        return $delegate($request);
+                $state->setSessionWasChanged($subsiteId === $originalSubsiteId);
```
tractorcow: You need to set this state before $delegate, otherwise it'll never be accessible to the controller.

tractorcow: You may need to change this to be a calculated property instead of an explicitly set one, that way any subsequent call to setSubsiteID() can trigger this to be flagged true.

robbieaverill:
65f85fa fixes #295

code/State/SubsiteState.php
```diff
@@ -28,19 +39,57 @@ public function getSubsiteId()
...
+    }
+
+    /**
+     * Get whether the subsite ID has been changed during a request, based on the original and current IDs
+     *
+     * @return bool
+     */
+    public function getSubsiteIdWasChanged()
+    {
+        return $this->originalSubsiteId !== $this->getSubsiteId();
```
tractorcow: Nice one :D

Issue #290, opened 2017-08-29
SS4 compatibility tracker - Subsites 2.0


robbieaverill:
Issue to track progress for upgrading to SilverStripe 4 compatibility.

## Requirements

- [x] add `SilverStripe\Subsites` and `SilverStripe\Subsites\Tests` namespaces #289 
- [x] implement core module namespaces #289
- [x] upgrade changes in API #289
- [x] add `$table_name` to DataObjects #289
- [x] run PHPCS linter #291 
- [x] update CI tool configurations (Travis, Scrutinizer, Codecov.io) #291 
- [x] fix Behat tests, uncomment Travis configuration to run them #294 
- [x] move templates into namespaced locations #304
- [x] move translations to namespaced classes #305 and #306 (which needs to be merged up)
- [x] [review](https://github.com/silverstripe/silverstripe-subsites/pull/289#discussion_r135681441) SilverStripe upgrader tool exclusion rules, add as necessary #307
- [x] add `$segment` to BuildTasks #303 
- [x] move `code/` to `src/` #301 

## Bugs

- fix middleware db  bootstrapping https://github.com/silverstripe/silverstripe-subsites/issues/295
- Asset admin doesn't work the same way as in SS3, so we'll need to come up with a way of allowing the FileSubsites extension to either filter the file list, visually indicate that a file is not visible in a given subsite at that point, or add a dropdown to the details screen to switch subsite #296 
- Page type blacklisting doesn't work #297 
- DB not ready kills application execution #302

## Ideas/enhancements

- Add silverstripe-webpack module configuration for building styles
- Move to using a middleware for Subsites rather than global state modification and data query hacks?
- Replace old use of ORM with newer equivalents, e.g. manual SQL joins with `SQLSelect::leftJoin` etc
- ["move all the common logic from these *Subsites extensions into a base SubsitesExtension class; FileSubsites would then only contain file-specific overrides to this."](https://github.com/silverstripe/silverstripe-subsites/pull/289#discussion_r135679409)

tractorcow:
Added an issue to fix middleware

robbieaverill:
Added issues for the asset admin/subsite compatibility, and for the page blacklist not working.

Perhaps once #294 is merged we should log the remaining tasks as separate issues and close this?

robbieaverill:
The PRs listed above are the last of the smaller fixes for this ticket. Once merged I'll close this and we can work from individual issues.

PR #304, opened 2017-09-07
API Move templates to namespaced location, remove unused templates

robbieaverill:
templates/LeftAndMain_Menu.ss looks to be a copy of the fully namespaced version, which is no longer used - removed. GridFieldAddFromTemplateButton also looks to be unused, so was also removed.

Issue #290


tractorcow:
@robbieaverill does the subsite menu appear ok without the overridden template?

robbieaverill:
@tractorcow there's actually two LeftAndMain_Menu.ss templates, one is namespaced correctly (and works) and the other isn't and doesn't get used. I guess it's a hangover from early upgrading efforts.

PR #305, opened 2017-09-07
NEW Namespace translations, implement into class localisation calls

robbieaverill:
Issue #290

@tractorcow do we need to freeze or fork translations for Subsites 3.x compatibility?


tractorcow:
@robbieaverill that's what I would do; You'd need to refresh all localisations from transifex, rename the keys (using upgrader), then push up and replace all master keys. To freeze localisations just remove `.tx` folder from all older branches.

robbieaverill:
Ok, thanks. `tx pull` didn't do anything, I guess the translations are fairly recently updated from the CWP recipe release. I'll add another PR now to remove the configuration from 3.x compat branches.

wilr:
Rebased, pushed and waiting for test review.

PR #310, opened 2017-09-11
FIX Page type blacklist not matching classes correctly (Fixes 297)

wilr:
Tidied up the UI, removed custom javascript in favour of core Toggle field.


robbieaverill:
Fix looks good, merge on green

PR #315, opened 2017-09-19
FIX Update XHR controller hook selector and refactor getCMSFields to use scaffolding and separate tabs

robbieaverill:
This pull request does:

* Use default scaffolding for getCMSFields on Subsite class
* Remove previously not shown tabs and form fields due to default scaffolding
* Re-arrange CMS fields for UX consistency
* Move Domains GridField in Subsite CMS fields into its own tab
* Fix the PJAX fragment reloading for the subsite list in the CMS when certain actions (e.g. adding a subsite) are performed
* Moves Subsite's private statics to the top of the class (from the middle) so they're with their brothers and sisters

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -35,7 +35,6 @@ use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
 use SilverStripe\Subsites\State\SubsiteState;
 use SilverStripe\Versioned\Versioned;
-
 use UnexpectedValueException;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Model/Subsite.php
```diff
@@ -652,78 +651,83 @@ public function canEdit($member = false)
...
-            $this->fieldLabel('Language'),
-            Injector::inst()->get(IntlLocales::class)->getLocales()
-        );
+            // Remove the default scaffolded blacklist field, we replace it with a checkbox set field
+            // in a wrapper further down. The RedirectURL field is currently not in use.
+            $fields->removeByName(['PageTypeBlacklist', 'RedirectURL']);
+
+            // Add the heading to the top of the fields list
+            $fields->fieldByName('Root.Main')
+                ->unshift(HeaderField::create('ConfigForSubsiteHeaderField', 'Subsite Configuration'));
```
fspringveldt: Header label not translatable. Should it be?

robbieaverill: Yep, thanks

PR #321, opened 2017-10-05
Implement Resettable interface on SubsiteState so it gets reset between unit tests

mikenz:


tractorcow:
All singletons should have been reset due to injector being unnested and nested again. Does this not happen?

mikenz:
@tractorcow It was causing test failures for me. A test left the current subsite ID as 2. The next test was able to see that.

Issue #298, opened 2017-09-06
SubsitesTreeDropdownField to be brought up to date with changes to TreeDropdownField


mikenz:
Since the React changes to TreeDropdownField, the javascript "API"s used by SubsitesTreeDropdownField no longer work.
https://github.com/silverstripe/silverstripe-subsites/blob/master/javascript/SubsitesTreeDropdownField.js#L12

https://github.com/silverstripe/silverstripe-admin/issues/142

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-admin/pull/414 (required from core first)
- [ ] https://github.com/silverstripe/silverstripe-subsites/pull/337

mikenz:
@wilr Here's what I ended up with:

SubsitesTreeDropdownField.js
```
(function ($) {
    $.entwine('ss', function ($) {
        /**
         * Choose a subsite from which to select pages.
         * Needs to clear tree dropdowns in case selection is changed.
         */
        $('select.subsitestreedropdownfield-chooser').entwine({
            onchange: function () {
                // TODO Data binding between two fields
                const name = this.attr('name').replace('_SubsiteID', '');
                let field = $('#Form_EditForm_' + name).first();
                field.setValue(0);
                field.refresh();
                field.trigger('change');
            }
        });

        /**
         * Add selected subsite from separate dropdown to the request parameters
         * before asking for the tree.
         */
        $('.TreeDropdownField.SubsitesTreeDropdownField').entwine({
            getAttributes() {
                const fieldName = this.attr('id').replace('Form_EditForm_', '');
                const subsiteID = $('#Form_EditForm_' + fieldName + '_SubsiteID option:selected').val();

                let attributes = this._super();
                attributes.data.urlTree += "?" + fieldName + "_SubsiteID=" + subsiteID;
                attributes.data.cacheKey = attributes.data.cacheKey.substring(0, 19) + "_" + subsiteID;
                return attributes;
            }
        });
    });
})(jQuery);
```

And SubsitesTreeDropdownField.php
```
...
    public function tree(HTTPRequest $request)
    {
        $session = Controller::curr()->getRequest()->getSession();
        $oldSubsiteID = Subsite::currentSubsiteID();
        if ($request->getVar($this->name . '_SubsiteID')) {
            $this->subsiteID = $request->getVar($this->name . '_SubsiteID');
        }
        $session->set('SubsiteID', $this->subsiteID);

        $results = parent::tree($request);

[...]

PR #322, opened 2017-10-05
SubsitesTreeDropdownField: Get the tree in the context of the correct subsite

mikenz:


mikenz:
I also realised it's a bit pointless until I raise a PR for https://github.com/silverstripe/silverstripe-subsites/issues/298 too.

PR #326, opened 2017-12-04
NEW Convert to vendor module

robbieaverill:


dhensby:
that looked relatively painless :)

PR #335, opened 2018-01-18
API Convert most of Subsite public statics to config properties

### LINKED ISSUE #174: Use config.yml
Is there a reason subsites doesn't make use of the Config class?

I just spent a couple of hours trying to work out why my files were getting a SubsiteID applied, then I realised that FileSubsites::$default_root_folders_global can not be set in a yml file. Doh!

robbieaverill:
I've left `Subsite::$force_subsite` as a public static but deprecated it in favour of using the SubsiteState instead, and have also left `$disable_subsite_filter` the way it is because it's used in a few other modules (including the CMS). Ideally in the next major release this will become part of the SubsiteState as well.

Resolves #174


robbieaverill:
Rebased

PR #345, opened 2018-02-22
FIX include action when creating absolute link

PsukheDelos:
Hi all!

I can see this issue has been fixed in master, but was hoping to get it into the SS3 version. 😄Any thoughts? 

Happy to provide more detail if necessary.

-G

### DIFF
--- a/code/extensions/SiteTreeSubsites.php
+++ b/code/extensions/SiteTreeSubsites.php
@@ -271,11 +271,15 @@ class SiteTreeSubsites extends DataExtension
         }
     }
 
-    public function alternateAbsoluteLink()
+	/**
+	 * @param null $action
+	 * @return string
+	 */
+    public function alternateAbsoluteLink($action = null)
     {
         // Generate the existing absolute URL and replace the domain with the subsite domain.
         // This helps deal with Link() returning an absolute URL.
-        $url = Director::absoluteURL($this->owner->Link());
+        $url = Director::absoluteURL($this->owner->Link($action));
         if ($this->owner->SubsiteID) {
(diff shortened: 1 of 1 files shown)


robbieaverill:
This does break backwards compatibility - see example: https://eval.in/961337

NightJar:
Hmm yeah, fair enough.
But; https://eval.in/961339

Since it's a `^3` only module, and SS3's support is for php5 only (until 3.6)...
Still, we support php7 so... ¯\\_(ツ)_/¯

robbieaverill:
>Since it's a ^3 only module, and SS3's support is for php5 only...
...Except we support php7 from 3.6, so... ¯\(ツ)/¯

Yep, we support PHP 5.3 to PHP 7.1

PsukheDelos:
@robbieaverill @NightJar What if I were to introduce a function called `getAbsoluteActionLink ($action = null)`? Or is that adding additional technical debt?

-G

NightJar:
G,
At this point I think your best bet would be to rather not add the parameter to the signature, but add a hook point later before returning from the function:

```
$this->extend('updateAlternateAbsoluteLink', &$url);
return $link;
```

This way you're not causing errors for folks living in the future, nor the past, and can modify the result until your heart's content via a `DataExtension`;
e.g.

```
return Controller::join_links($link, $action);
```
(which is really all that happens when passing an `$action` param anyway)

PsukheDelos:
Thanks @NightJar! I will close this off now. 😄

NightJar:
Of course, you might have a touch of trouble in that this is already a `DataExtension` it seems...
But the idea is still solid.

robbieaverill:
The other option is that we can just break semver for the sake of doing this method properly

PR #353, opened 2018-03-26
FIX be consistent with the variable setting

NightJar:
So as to avoid fatal errors while attempting to count.

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -15,9 +15,11 @@ use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\FormAction;
 use SilverStripe\Forms\ToggleCompositeField;
 use SilverStripe\i18n\i18n;
+use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataExtension;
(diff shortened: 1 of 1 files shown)


robbieaverill:
The test failure is unrelated, but needs to be fixed separately. We should break the assertion up so it doesn’t fail when the stage is present in the query string (SS 4.2 change)

src/Extensions/SiteTreeSubsites.php
```diff
@@ -106,7 +108,7 @@ public function onBeforeWrite()
     public function updateCMSFields(FieldList $fields)
     {
         $subsites = Subsite::accessible_sites('CMS_ACCESS_CMSMain');
-        $subsitesMap = [];
+        $subsitesMap = new Map(ArrayList::create());
```
dhensby: It may be better to put this into an else as we'd be using unnecessary memory to construct two needless objects here if there are subsites available.

dhensby:
broken tests fixed here: https://github.com/silverstripe/silverstripe-subsites/pull/354

dhensby:
I've rebased this and fixed an indentation issue, hopefully we'll see green :)

PR #359, opened 2018-04-20
FIX Automatically create default SiteTree records for new subsites

### LINKED ISSUE #252: New subsite should automatically get a "home" page (infinite redirects otherwise)
Every SilverStripe website needs a page with the `home` URLSegment. Creating an empty subsite leads to infinite redirects when not copying structure from a template. This is hardly ideal default behaviour.

robbieaverill:
Resolves #252 (for SS3). For SS4 we can modify the logic slightly:

```php
SubsiteState::singleton()->withState(function (SubsiteState $newState) {
    $newState->setSubsiteId($this->ID);

    DB::quiet();
    (new SiteTree)->requireDefaultRecords();
});
```

^ in unit tests and `Subsite::createDefaultPages`.

This will create a home page at least. The logic in `SiteTree::requireDefaultRecords` uses manual SQL queries to determine whether an about us or contact page exist, so that's unlikely to work here. We get a home page at least, which was the original requirement.

### DIFF
--- a/code/model/Subsite.php
+++ b/code/model/Subsite.php
@@ -743,15 +743,34 @@ class Subsite extends DataObject
     }
 
     /**
-     * Whenever a Subsite is written, rewrite the hostmap
+     * Whenever a Subsite is written, rewrite the hostmap and create some default pages
      *
      * @return void
      */
     public function onAfterWrite()
     {
         Subsite::writeHostMap();
+        if ($this->isChanged('ID')) {
+			$this->createDefaultPages();
+		}
         parent::onAfterWrite();
     }
+
+	/**
+	 * Automatically create default pages for new subsites
+	 */
+    protected function createDefaultPages()
+	{
+		$currentSubsite = Subsite::currentSubsiteID();
+		Subsite::changeSubsite($this->ID);
+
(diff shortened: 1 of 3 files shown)


code/model/Subsite.php
```diff
@@ -750,9 +758,41 @@ public function validate()
     public function onAfterWrite()
     {
         Subsite::writeHostMap();
+        $this->createDefaultPages();
```
NightJar: I think this can be done via `if ($this->isChanged('ID')) {` which would avoid the need for the new member property and the abstraction to the new function (although a protected function isn't really the worst thing). If this remains true (I'm 90% certain this was the case in SS3 at least) then it would make the code more readable and easier to follow.

robbieaverill: It would, but I wouldn't be at all surprised if this doesn't work. The last time I tried to track field level changes during a write I had to do this individually for every field I wanted to track because the framework calls ->write() multiple times. Nevermind, let's change the API and see if the test still passes!

robbieaverill: Works like a charm! Thanks for the suggestion 👍

PR #361, opened 2018-06-01
FIX Do not make subsite based file permission decisions when no subsite is set

robbieaverill:
Fixes the two assets permission tests that are noted as failing in the CWP 2.0 changelog, and mentioned here: https://github.com/silverstripe/silverstripe-assets/issues/136


robbieaverill:
Note SS 4.2.x build is broken externally at the moment

PR #362, opened 2018-06-01
FIX Use correct table name for Group model when performing DB upgrades from older versions

### LINKED ISSUE #356: DB error when updating from old DB (2.3)
I just encountered this error while updating a really old site to latest Subsites in SS4.1:

> ERROR [Emergency]: Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query:
> 
> SHOW FULL FIELDS IN "SilverStripe\Security\Group"
> 
> Table 'database.SilverStripe\Security\Group' doesn't exist

It's caused by `SilverStripe\Subsites\Extensions\GroupSubsites->requireDefaultRecords`

robbieaverill:
Fixes #356


robbieaverill:
Broken tests in SS 4.2.x and 4.x are fixed with silverstripe/silverstripe-installer#230

PR #363, opened 2018-06-01
FIX Re-enable Behat using chromedriver and silverstripe/recipe-testing

### LINKED ISSUE #323: Fix Behat build in Travis
A while back a change in core meant the subsites Behat tests started failing, so we've added them to allowed failures.

We should revisit and ensure they're working.

robbieaverill:
Fixes #323

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
   # Init PHP
   - phpenv rehash
   - phpenv config-rm xdebug.ini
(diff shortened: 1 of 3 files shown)


robbieaverill:
Broken tests in SS 4.2.x and 4.x are fixed with https://github.com/silverstripe/silverstripe-installer/pull/230

robbieaverill:
We need to sort out memory issues with the coverage runs now... =/

PR #364, opened 2018-06-01
if the table names are not case sensitive this condition always thing…

fonsekaean:
…s the table is hasn't been created.

We had this issue on a window XAMPP / WAMPP installation where mysql was not case sensitive, and the subsites were always ignored. 

This modification checks the lower case tables names against a lowercase array

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -276,7 +276,7 @@ class Subsite extends DataObject
 
             /** @skipUpgrade */
             $domainTableName = $schema->tableName(SubsiteDomain::class);
-            if (!in_array($domainTableName, DB::table_list())) {
+            if (!in_array(strtolower($domainTableName), array_map("strtolower", DB::table_list())) ) { // case sensitive table names
                 // Table hasn't been created yet. Might be a dev/build, skip.
                 return 0;
             }


fonsekaean:
I agree about the case sensitivity is something which has to be configured in the DB level. However thats something which can be missed quite easily, and also the CMS doesnt force it anyway too. 

Thanks for approving the pull request.

dhensby:
Well, core should be renaming table case if it's wrong... so this can only be a problem if you've got a mis-configured DB server. I'm not sure we should be putting such a specific fix into a module when the core framework would still be susceptible to this problem.

src/Model/Subsite.php
```diff
@@ -276,7 +276,7 @@ public static function getSubsiteIDForDomain($host = null, $checkPermissions = t
 
             /** @skipUpgrade */
             $domainTableName = $schema->tableName(SubsiteDomain::class);
-            if (!in_array($domainTableName, DB::table_list())) {
+            if (!in_array(strtolower($domainTableName), array_map("strtolower", DB::table_list())) ) { // case sensitive table names
```
NightJar: This should be either
`!array_key_exists(strtolower($domainTableName))` (ideally)
or
`!(array_key_exists($domainTableName)  && in_array($domainTableName))`
as per the [return value](https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/Connect/DBSchemaManager.php#L969) of `tableName` being keyed in the format of `[strtolower($name) => $name]`

NightJar:
@dhensby raises a good point though, this is probably better approached by checking if there's some kind of build going on, and/or using the same method core does to check if a table already exists (i.e. whether or not to `CREATE TABLE` in a dev/build) - since the only reason for this check is ultimately prevent an error if we are building.

Personally I don't see why the model should need to know this information anyway, surely the relationship should to `SubsiteDomain` should be enough - but I don't have enough experience with the code of this module to make a fully formed opinion.

PR #369, opened 2018-06-19
FIX Do no provide input for canEdit or canPublish if no subsites exist

robbieaverill:


src/Extensions/SiteTreeSubsites.php
```diff
@@ -273,15 +273,22 @@ public function alternateSiteConfig()
...
+     * @return bool|null
      */
     public function canEdit($member = null)
     {
         if (!$member) {
             $member = Security::getCurrentUser();
         }
 
+        // Do not provide any input if there are no subsites configured
+        if (!Subsite::get()->count()) {
```
dhensby: `exists` would be better here

robbieaverill: Sure, follow up at #371

PR #374, opened 2018-07-04
FIX apply SubsiteID getVar to CMS Preview fetches

### LINKED ISSUE #263: Preview panel breaking /admin section on SubSite when viewing Mainsite
If you access the /admin section of a subsite, then change the Site to "Main Site". When trying to edit a page the preview panel causes the CMS to redirect to a broken page.

Replication:
- Login to /admin of subsite
- Change site being edited to "Main Site"
- Select a page to edit on the Main Site.

NightJar:
A preview link must be loaded on the same domain the CMS is loaded
through, which was previously causing issues when a page (identified
via URLSegment) did not exist on the subsite domain. By _always_
including the identifier in the preview link, this should never happen.

Resolves #263

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -382,17 +382,31 @@ class SiteTreeSubsites extends DataExtension
 
     /**
      * Use the CMS domain for iframed CMS previews to prevent single-origin violations
-     * and SSL cert problems.
-     * @param null $action
+     * and SSL cert problems. Always set SubsiteID to avoid errors because a page doesn't
+     * exist on the CMS domain.
+     *
+     * @param string &$link
+     * @param string|null $action
+     * @return string
+     */
+    public function updatePreviewLink(&$link, $action = null)
+    {
+        $url = Director::absoluteURL($this->owner->Link($action));
+        $link = HTTP::setGetVar('SubsiteID', $this->owner->SubsiteID, $url);
+        return $link;
+    }
+
+    /**
+     * This function is marked as deprecated for removal in 5.0.0 in silverstripe/cms
+     * so now simply passes execution to where the functionality exists for backwards compatiblity.
+     *
+     * @param string|null $action
      * @return string
      */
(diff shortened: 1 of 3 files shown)


src/Extensions/SiteTreeSubsites.php
```diff
@@ -382,17 +382,16 @@ public function alternateAbsoluteLink($action = null)
 
     /**
      * Use the CMS domain for iframed CMS previews to prevent single-origin violations
-     * and SSL cert problems.
+     * and SSL cert problems. Always set SubsiteID to avoid errors because a page doesn't
+     * exist on the CMS domain.
      * @param null $action
      * @return string
      */
-    public function alternatePreviewLink($action = null)
```
robbieaverill: This would be a breaking change. How much do you care about it? Looks like you need to change the arguments too. You could either do that in a non breaking way (using `func_get_args`) or create a new method and deprecate the old method. Would be good to get this out without a major version bump though!

NightJar: Hmm, a good (overlooked) point. Will fix.
The reasoning for the original change is [deprecation](https://github.com/silverstripe/silverstripe-cms/blob/4.0.0/code/Model/SiteTree.php#L560)

robbieaverill: I see your deprecation and raise you with... more deprecation

src/Extensions/SiteTreeSubsites.php
```diff
@@ -382,17 +382,30 @@ public function alternateAbsoluteLink($action = null)
...
+     *
+     * @param string &$link
+     * @param string|null $action
+     * @return string
+     */
+    public function updatePreviewLink(&$link, $action = null)
+    {
+        $url = Director::absoluteURL($this->owner->Link($action));
+        $link = HTTP::setGetVar('SubsiteID', $this->owner->SubsiteID, $url);
+        return $link;
```
robbieaverill: Don’t think this needs to return anything but it probably doesn’t matter

PR #381, opened 2018-08-16
allow disabling filter using queryParam

lekoala:
It is much better than relying on global state

### DIFF
--- a/src/Extensions/FileSubsites.php
+++ b/src/Extensions/FileSubsites.php
@@ -52,6 +52,9 @@ class FileSubsites extends DataExtension
         if (Subsite::$disable_subsite_filter) {
             return;
         }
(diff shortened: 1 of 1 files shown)


dhensby:
what's the difference between this and #382?

robbieaverill:
One is for FileSubsites and one for GroupSubsites. I think since the existing tests pass then it's OK, but I feel like there's something more that this change needs, e.g. if you're replacing some global state filtering there should be some code removed too?

lekoala:
Well, the idea is to have a consistent way to disable subsite filtering using query params. sorry for using two pr for that (just much easier to do so right from github and i've been lazy)

i won't remove anything for now to keep compatibility, also sometimes you cannot easily access the query itself depending on the context.

robbieaverill:
Thanks =)

PR #387, opened 2018-08-23
FIX Pages now correctly duplicate children across subsites

### LINKED ISSUE #384: "Include children pages" when copying to subsite does not work
It seems that the "Include children pages" option when copying a page to a subsite does not do anything.

Within the `copytosubsite` function in `LeftAndMainSubsites.php` there is this code:

`$newPage = $page->duplicateToSubsite($subsite->ID, $includeChildren);`

However, the `duplicateToSubsite` function within `SiteTreeSubsites.php` only accepts one argument:

`public function duplicateToSubsite($subsiteID = null)`

It looks like the ability to copy children pages has vanished? I've tested this feature and it definitely only copies the parent page across.

robbieaverill:
Resolves #384 

I've removed some of the custom duplication (and legacy) logic and replaced it with some more "core" API calls inside a modified subsite state callback, and have restored the use of `$includeChildren` in recursive duplication of children.

#386 deprecates the `duplicateSubsiteRelations` method since we should now use cascade_duplicates config which will automatically inherit the subsite state now with this PR change (see `onBeforeDuplicate`)

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\Subsites\Extensions;
 
-use Page;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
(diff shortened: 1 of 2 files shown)


ScopeyNZ:
Thanks for the contribution @robbieaverill ! 🎉

PR #398, opened 2018-10-19
FIX Ignore ports when matching domain for subsite

### LINKED ISSUE #373: Specifying a "Domain" for a Subsite does not account for Ports
Or rather _always_ accounts for ports - which is counter intuitive to the "domain" moniker.
Being web developers we're all aware that a URL is destructured into components, some of which are optional;
`scheme:[//[user:pass@]host[.domain][:port]]path[?query][#fragment]`

image

The entry box for the Domain segment specifically directs us to leave out the `scheme` section (listed as 'protocol'). The assumption then made is that the Domain input field is looking for the `host[.domain]` section of the URL...
However as my site is serving on a port other than the default standard `80`, I was surprised to learn that the port is also **mandatory** here in order to have the subsite serve itself on the separate sub-domain created for it.

**What:**
A site is served from `localhost`
A subsite is configured for the domain `subby.localhost`
The web server is configured to listen on port `8080`
**Expected**
`localhost:8080` will serve the main site.
`subby.localhost:8080` will serve the subsite.
**Actual**
`localhost:8080` serves the main site.
`subby.localhost:8080` _also_ serves the main site.

**Resolution**
I see three options to resolve this:
- Ports are ignored in matching a subsite (possibly as a fallback to no exact match)
- Options are added to strictly match port or fall back to excluding it (such as the existing options for scheme/protocol)
- The Domain input field description text is updated to indicate a port may also be needed (along side the current protocol directive).

robbieaverill:
Fixes #373

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -257,6 +257,10 @@ class Subsite extends DataObject
             $host = $_SERVER['HTTP_HOST'];
         }
 
+        // Remove ports, we aren't concerned with them in terms of detecting subsites via domains
+        $hostParts = explode(':', $host);
+        $host = reset($hostParts);
+
         $matchingDomains = null;
         $cacheKey = null;
         if ($host) {
--- a/tests/php/SubsiteTest.php
+++ b/tests/php/SubsiteTest.php
@@ -193,9 +193,20 @@ class SubsiteTest extends BaseSubsiteTest
         );
         $this->assertEquals(
             $subsite1->ID,
+            Subsite::getSubsiteIDForDomain('example.org:1123'),
+            'Exact matches without strict checking when not using www prefix and ignores port'
+        );
+        $this->assertEquals(
+            $subsite1->ID,
             Subsite::getSubsiteIDForDomain('www.example.org'),
             'Matches without strict checking when using www prefix, '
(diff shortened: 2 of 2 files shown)


NightJar:
Ideal, thanks :)

src/Model/Subsite.php
```diff
@@ -257,6 +257,10 @@ public static function getSubsiteIDForDomain($host = null, $checkPermissions = t
             $host = $_SERVER['HTTP_HOST'];
         }
 
+        // Remove ports, we aren't concerned with them in terms of detecting subsites via domains
+        $hostParts = explode(':', $host);
```
dhensby: `explode(':', $host, 2)`

NightJar: I'm not sure this is necessary. What case do you see it being of benefit? If someone specifies something invalid?
It could be pointed out that `user:pass@domain:port` has 2 colons I suppose...

dhensby: Well, it's faster if the internals know to stop scanning the string after the first one is found, and we use reset to get the first part later in the function so the invalid point is moot

robbieaverill: Patched in cd71c2b @dhensby

Issue #357, opened 2018-04-13
Content Editor cannot create link in WYSIWYG editor (nor anchor / email / file)


baukezwaan:
## Affected Version

```
phpunit/phpunit            5.7.27             The PHP Unit Testing framework.
silverstripe-themes/simple dev-master 4d546a4 The SilverStripe simple theme (default SilverStripe 3 theme)
silverstripe/recipe-cms    1.0.3              SilverStripe recipe for fully featured page and asset content editing
silverstripe/recipe-plugin 1.2.0              Helper plugin to install SilverStripe recipes
silverstripe/subsites      2.0.1              Run multiple sites from a single SilverStripe install.
```

Issue appears to be on SilverStripe CMS version 4.0.3 and 4.1

## Description

A user that belongs to a group without 'Access to all CMS sections' is getting an error like
`Unexpected token < in JSON at position 0`
when you try to insert a link or anchor into the HTMLeditor on a page.

![knipsel](https://user-images.githubusercontent.com/795238/38736820-91b4753e-3f2d-11e8-9dc4-325c71056923.PNG)

## Steps to Reproduce

On a clean SilverStripe installation with Subsites, create a user that belongs to the Content Author group. Then try to insert a link into a page.

baukezwaan:
A little bit of experimenting with xdebug seems to point to LeftAndMainSubsites->canAccess. 

It checks to see if the current Member has ADMIN or CMS_ACCESS_LeftAndMain. My Member doesn't have these rights, but have only access to certain parts of the CMS. 
Additionally, canAccess check if we have access to current section on the current Subsite, but that doesn't seem to account for the possibility that we are trying to access an action on LeftAndMain directly. 

I don't know what would be a proper fix, but when I insert an if( $oController instanceof LeftAndMain ) return true; I at least can access the TinyMCE modals.

adrexia:
This appears to be fixed in 6af9854, but breaks again after that commit was reverted. Does anyone have more context on why that commit was reverted?

danaenz:
Hi all,

We've had this issue reported as well. Affected versions:
silverstripe/recipe-cms 1.1.3
 silverstripe/subsites 2.2.0
silverstripe/framework 4.1.3
silverstripe/cms 4.1.3

Since I need to fix this urgently for our client I'm attempting to find a workaround rather than roll back. I'm unsure which version I should roll back to if I need to, perhaps 1.1.0?

ScopeyNZ:
[...]

PR #408, opened 2019-02-10
FIX Content editor group users can now insert links into contents while using subsites

robbieaverill:
This change allows users with CMSMain permissions to have access in LeftAndMainSubsites::canAccess(), since `ModalController` doesn't define its own permission codes, and usually modals are related to CMS content.

Fixes https://github.com/silverstripe/silverstripe-subsites/issues/357

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,11 @@ matrix:
     - php: 7.0
       env: DB=PGSQL INSTALLER_VERSION=4.1.x-dev PHPUNIT_TEST=1
     - php: 7.1
-      env: DB=MYSQL INSTALLER_VERSION=4.1.x-dev BEHAT_TEST=1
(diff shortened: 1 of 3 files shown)


robbieaverill:
Ok so the "Insert Link" label was changed to "Insert link" in SilverStripe 4.3 (see test fixes: https://github.com/silverstripe/silverstripe-cms/commit/41c0b8fb85b7ac11f18eb1813f9e063e13cbafa2), which is what caused the Behat failures on SilverStripe 4.1. I've moved the Behat builds to run on 4.3 or newer, since we'll stop supporting 4.1 sooner than 4.3, but the test failure in 4.1 is isolated to tests - it works in practice.

I chose to do this rather than modify the behat-extension module to support case insensitive xpath lookups.

PR #413, opened 2019-05-17
FIX Adding an extension to ensure blacklisted page types are disallowed

### LINKED ISSUE #351: Cannot limit page types for a subsite
**Affected versions**:

* 2.0.x-dev using SS 4.1.x-dev

Still passing on 2.0.x-dev using SS 4.0.x-dev

When I select a subsite in admin and disallow an Error Page for example, switch to that subsite and go to add a new page, the Error Page is still a valid option and I can still add it.

ScopeyNZ:
Fixes #351

In my testing it seemed I was denied from creating pages with `canCreate` but you still got a full list of page types regardless of the subsite blacklist. This PR adds an extension to limit the list. This might be getting a bit close to the line of "too many impacting changes for a patch" - happy to retarget or alter the PR.

### DIFF
--- a/_config/extensions.yml
+++ b/_config/extensions.yml
@@ -14,6 +14,7 @@ SilverStripe\CMS\Controllers\ContentController:
 SilverStripe\CMS\Controllers\CMSPageAddController:
   extensions:
     - SilverStripe\Subsites\Extensions\CMSPageAddControllerExtension
+    - SilverStripe\Subsites\Extensions\PageAddBlacklistExtension
 
 SilverStripe\Admin\LeftAndMain:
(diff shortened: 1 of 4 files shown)


ScopeyNZ:
I've just realised that it probably was intended to be the way it was... The page types just weren't getting disabled...

robbieaverill:
>I've just realised that it probably was intended to be the way it was... The page types just weren't getting disabled...

That's right. The page types are supposed to be there but be disabled.

ScopeyNZ:
Oh well I wasted an hour. Close this then 😅

Issue #251, opened 2016-06-12
"Protocol" in subsite domain editor should have a default value


chillu:
![image](https://cloud.githubusercontent.com/assets/111025/15994104/0539847a-314f-11e6-8bcc-d98a9829ff7e.png)

robbieaverill:
Still an issue in SS4

robbieaverill:
Let's just call this a bug =D

robbieaverill:
PR to fix this and a duplicate description bug here https://github.com/silverstripe/silverstripe-subsites/pull/417.

Duplicate description issue:
![image](https://user-images.githubusercontent.com/5170590/58667095-3685b000-8389-11e9-8bda-313bf24f51f0.png)

PR #417, opened 2019-05-30
FIX Domains now default to "Automatic" protocol, and have the correct help description

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-subsites/issues/251 and an issue where the "Mark this as the default domain for this subsite" description was also shown on the protocol option set as well as the "IsPrimary "checkbox


robbieaverill:
Re-targetted at 2.2 because 2.3 requires SilverStripe 4.4

PR #419, opened 2019-05-30
FIX allowed pagetypes displaying incorrectly when switching subsite

### LINKED ISSUE #351: Cannot limit page types for a subsite

Cheddam:
Introduces an extension to add the current Subsite ID to the SiteTree Hints Cache Key so that the allowed pagetypes are correctly cached on a per-Subsite basis. Resolves #351.

This patch depends on [an update to the CMS module](https://github.com/silverstripe/silverstripe-cms/pull/2442) that provides this extension point. The code is inert when matched with existing CMS versions.

### DIFF
--- a/_config/extensions.yml
+++ b/_config/extensions.yml
@@ -50,6 +50,7 @@ SilverStripe\Admin\SecurityAdmin:
 
 SilverStripe\CMS\Controllers\CMSMain:
   extensions:
+    - SilverStripe\Subsites\Extensions\HintsCacheKeyExtension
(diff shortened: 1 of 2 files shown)


robbieaverill:
Want to retarget at 2.3 branch?

PR #421, opened 2019-05-30
FIX Field labels for subsites virtual pages are no longer repeated

robbieaverill:
They were all showing "Title"


robbieaverill:
Re-targetted at 2.2 because 2.3 requires SilverStripe 4.4

Issue #423, opened 2019-05-30
Cannot save a virtual page after initially creating it with fluent installed


robbieaverill:
`validURLSegment()` throws a conflict error where URLSegment is ambiguous if you have fluent installed, e.g. in CWP

Found while testing https://github.com/silverstripe/silverstripe-subsites/issues/185

robbieaverill:
Fixed in https://github.com/silverstripe/silverstripe-subsites/pull/425

PR #425, opened 2019-05-30
FIX Subsites virtual pages now allow you to re-save them when used in conjunction with silverstripe-fluent

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-subsites/issues/423

URLSegment from custom SQL filter was ambiguous, using the ORM fixes this

### DIFF
--- a/src/Pages/SubsitesVirtualPage.php
+++ b/src/Pages/SubsitesVirtualPage.php
@@ -8,7 +8,6 @@ use SilverStripe\CMS\Model\VirtualPage;
 use SilverStripe\Control\Controller;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Forms\DropdownField;
-use SilverStripe\Forms\LabelField;
 use SilverStripe\Forms\LiteralField;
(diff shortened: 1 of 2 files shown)


src/Pages/SubsitesVirtualPage.php
```diff
@@ -223,33 +222,27 @@ public function validURLSegment()
...
+
         // Veto the validation rules if its false. In this case, some logic
         // needs to be duplicated from parent to find out the exact reason the validation failed.
         if (!$isValid) {
-            $IDFilter = $this->ID ? "AND \"SiteTree\".\"ID\" <> $this->ID" : null;
-            $parentFilter = null;
+            // Exclude the current page from the filter
+            if ($this->exists()) {
+                $filters['ID:not'] = $this->ID;
+            }
```
unclecheese: Minor point, but without this check, isn't `$this->ID` a `0` anyway? In which case `ID:not => 0` is perfectly fine?

robbieaverill: True, will update

PR #427, opened 2019-05-31
FIX Improving support for cascading themes

### LINKED ISSUE #392: Cascading Themes?
Is it possible to use cascading themes with subsites? Or to define the allowed themes from yml? It looks like they might currently be able to be set from config.php, but it would be super handy if we could define a list of themes from yml. eg:

```
SilverStripe\Subsites\Model\Subsite:
  allowed_themes:
    - '/theme-default'
    - '/theme-other'
```

Or, if cascading needs to be defined: 
```
SilverStripe\Subsites\Model\Subsite:
  allowed_themes:
    defaultThemeName:
       - '$public'
       - '/theme-default'
       - '$default'
    otherThemeName:
       - '$public'
       - '/theme-other'
       - '$default'
```

ScopeyNZ:
- Fixes an issue where themes would cascade "up" the list of themes
- Provides configuration for defining custom theme options with their own sets of cascading themes

Fixes #392

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -28,6 +28,7 @@ use SilverStripe\Security\Group;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
+use SilverStripe\Subsites\Service\ThemeResolver;
 use SilverStripe\Subsites\State\SubsiteState;
 use SilverStripe\Versioned\Versioned;
 use UnexpectedValueException;
@@ -182,7 +183,6 @@ class Subsite extends DataObject
     /**
      * Gets the subsite currently set in the session.
      *
-     * @uses ControllerSubsites->controllerAugmentInit()
      * @return DataObject The current Subsite
      */
     public static function currentSubsite()
@@ -787,7 +787,7 @@ class Subsite extends DataObject
      */
     public function allowedThemes()
     {
-        if ($themes = self::$allowed_themes) {
+        if (($themes = self::$allowed_themes) || ($themes = ThemeResolver::singleton()->getCustomThemeOptions())) {
             return ArrayLib::valuekey($themes);
         }
 
--- a/src/Service/ThemeResolver.php
+++ b/src/Service/ThemeResolver.php
@@ -59,8 +59,11 @@ class ThemeResolver
         $index = array_search($siteTheme, $themes);
 
         if ($index > 0) {
+            // 4.0 didn't have support for themes in the public webroot
+            $publicConstantDefined = defined('SSViewer::PUBLIC_THEME');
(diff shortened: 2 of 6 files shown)


src/Service/ThemeResolver.php
```diff
@@ -0,0 +1,95 @@
...
+     *   'theme-1' => [
+     *     '$public',
+     *     'starter',
+     *     '$default',
+     *   ],
+     *   'theme-2' => [
+     *     'custom',
+     *     'watea',
+     *     'starter',
+     *     '$public',
```
robbieaverill: The public entry would always be at the start

ScopeyNZ: Eh? You can choose.

tests/php/Service/ThemeResolverTest.php
```diff
@@ -0,0 +1,143 @@
...
+    public function testSubsiteWithCustomThemeDoesNotCascadeUpTheList()
+    {
+        $subsite = new Subsite();
+        $subsite->Theme = 'main';
+
+        $resolver = new ThemeResolver();
+
+        $expected = [
+            'main', // 'main' is moved to the top
+            SSViewer::PUBLIC_THEME, // $public is preserved
```
robbieaverill: I'm not sure this is a realistic use case - public is put first in theme lists

ScopeyNZ: This is how it works currently - the chosen theme is moved _above_ public. It'd technically be breaking if I changed this.

src/Model/Subsite.php
```diff
@@ -783,7 +783,7 @@ public function fieldLabels($includerelations = true)
      */
     public function allowedThemes()
     {
-        if ($themes = self::$allowed_themes) {
+        if (($themes = self::$allowed_themes) || ($themes = ThemeResolver::singleton()->getCustomThemeOptions())) {
```
robbieaverill: Do you want to make a separate PR to master to deprecate `$allowed_themes`?

ScopeyNZ: Yes - I was going to deprecate this and another useless extension point after I merge up to `2.3`.

robbieaverill:
Merge on green

robbieaverill:
Ah, the test failure is valid - the public theme constant was not available in 4.0

ScopeyNZ:
Ok there's still stuff to fix with that `$public` stuff. I'll sort it out later. I think it's important that it's handled consistently (compared to previously) with the new code - but I'll have to figure out a way to incorporate that into tests that run with and without `$public` support.

ScopeyNZ:
I cheated by rebasing onto `2.1`. Hopefully tests are green after the latest commit.

PR #428, opened 2019-06-12
Detect domains correctly in Director sub-calls

nikrolls:
Previously it relied on the PHP-level `$_SERVER` variable; now it will use the `HTTPRequest` so it works correctly in more situations.


robbieaverill:
Looks like a good change! Thanks. Would you mind running a phpcbf linter over it? The test class needs a few PSR-12 tweaks

nikrolls:
Thanks @robbieaverill! 

Apologies, those were my default settings - I've applied the corrections from PHPCBF in Travis and pushed again (I force-pushed on automatic because I didn't expect you to have got to it so quickly!

nikrolls:
Thanks @robbieaverill!

Issue #437, opened 2019-07-15
Running dev/build via CLI throws [Notice]: Undefined index: HTTP_HOST


ChrissiQ:
Because `$_SERVER` variables are not set in the CLI, the `dev/build` command thows a notice:

````
[1;31mERROR [Notice]: Undefined index: HTTP_HOST
IN GET dev/build
[0m[31mLine 189 in [path]/vendor/silverstripe/subsites/src/Model/SubsiteDomain.php

[0mSource
======
  180:     /**
  181:      * Retrieves domain name with wildcards substituted with actual values
  182:      *
  183:      * @todo Refactor domains into separate wildcards / primary domains
  184:      *
  185:      * @return string
  186:      */
  187:     public function getSubstitutedDomain()
  188:     {
* 189:         $currentHost = $_SERVER['HTTP_HOST'];
  190: 
  191:         // If there are wildcards in the primary domain (not recommended), make some
  192:         // educated guesses about what to replace them with:
  193:         $domain = preg_replace('/\.\*$/', ".{$currentHost}", $this->Domain);
  194: 
  195:         // Default to "subsite." prefix for first wildcard

Trace
=====
Monolog\ErrorHandler->handleError(8, Undefined index: HTTP_HOST, [path]/vendor/silverstripe/subsites/src/Model/SubsiteDomain.php, 189, Array)
SubsiteDomain.php:189

SilverStripe\Subsites\Model\SubsiteDomain->getSubstitutedDomain()
ViewableData.php:133

SilverStripe\View\ViewableData->__get(SubstitutedDomain)
Subsite.php:874

SilverStripe\Subsites\Model\Subsite->domain()
ErrorPageSubsite.php:47

SilverStripe\Subsites\Extensions\ErrorPageSubsite->updateErrorFilename(error-404.html, 404, , , , , )
Extensible.php:468

SilverStripe\View\ViewableData->extend(updateErrorFilename, error-404.html, 404)
ErrorPage.php:399

SilverStripe\ErrorPage\ErrorPage::get_error_filename(404, SilverStripe\ErrorPage\ErrorPage)
ErrorPage.php:411

SilverStripe\ErrorPage\ErrorPage->getErrorFilename()
ErrorPage.php:268

SilverStripe\ErrorPage\ErrorPage->hasStaticPage()
ErrorPage.php:167

SilverStripe\ErrorPage\ErrorPage->requireDefaultRecordFixture(Array)
ErrorPage.php:140

[...]

PR #440, opened 2019-07-26
FIX Prevent undefined index notice when trying to determine HTTP_HOST during dev/build

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-subsites/issues/437

There's no host information available when run from command line because `CLIRequestBuilder` doesn't provide it, so we need to have a fallback to a sensible default.

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -870,7 +870,7 @@ class Subsite extends DataObject
         }
 
         // If there are no objects, default to the current hostname
-        return isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : 'localhost';
+        return Director::host();
     }
 
     /**
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -870,7 +870,7 @@ class Subsite extends DataObject
         }
 
         // If there are no objects, default to the current hostname
-        return $_SERVER['HTTP_HOST'];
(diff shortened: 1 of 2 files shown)


src/Model/Subsite.php
```diff
@@ -870,7 +870,7 @@ public function domain()
         }
 
         // If there are no objects, default to the current hostname
-        return $_SERVER['HTTP_HOST'];
+        return isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : 'localhost';
```
NightJar: Please use `Director::host()` instead of this ternary operation.

src/Model/SubsiteDomain.php
```diff
@@ -185,7 +185,7 @@ public function getFullProtocol()
      */
     public function getSubstitutedDomain()
     {
-        $currentHost = $_SERVER['HTTP_HOST'];
+        $currentHost = isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : 'localhost';
```
NightJar: Please use `Director::host()` instead of this ternary operation.

src/Model/Subsite.php
robbieaverill: Doing so would introduce many more logic paths which were previously not included. I think it makes sense to do it anyway though.

PR #443, opened 2019-10-09
NEW show alert when subsite changes in another tab

### LINKED ISSUE #314: Subsites not safe for multiple tabs
I'm not sure that there is an easy fix for this but we do believe that it has caused issues for us.

Since the ID of the current subsite is stored in the session this can cause problems if working on different sites across multiple tabs:

1. Open two tabs on main site
1. Switch to a subsite in tab 2
1. In tab 1 create a new page

Expected:

Page is created in main site.

Actual

Page is created in subsite.

It is easy to not notice this because initially it _looks_ like the page has be created in the main site as expected.

### LINKED ISSUE #411: "Simple" fix to make subsites multi-tab safe
Until #314 can be fixed fully, make subsites multi-tab safe as follows:

Refresh every "background tab" when you switch subsites in the active tab. This can likely be done through shared state (local storage) and web worker events or something. It's a stop gap measure, but a whole lot quicker to fix than making every single view and edit request in the CMS subsite aware through stateless payloads.

NightJar:
Adds a new feature that alerts a user who is editing a site in the same browser but over multiple tabs, and changes the subsite in one tab. Normally nothing untoward would happen until the user then tries to save data for a subsite that is no longer the current one in session. Now a javascript powered modal will pop up alerting users to the fact that the subsite has changed, and includes a button to revert in order to continue editing (so changes aren't lost).

Resolves #314 
Closes #411

Tested working in IE11.

image


client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
+/* jslint browser: true, nomen: true */
```
emteknetnz: seems like this file should have a .jsx extension

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+    },
+
+    showHideSubsiteList() {
+      $('#Form_ItemEditForm_Subsites').parent().parent().css('display', ($('#Form_ItemEditForm_AccessAllSubsites_1').is(':checked') ? 'none' : ''));
+    }
+  });
+
+  $('.cms-edit-form').entwine({
+    /**
+     * TODO: Fix with Entwine API extension. See https://github.com/silverstripe/silverstripe-subsites/pull/125
```
emteknetnz: is this TODO going to get done?

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+      if (!subsiteId) {
+        return;
+      }
+
+      // Inject the SubsiteID into internal links.
+      $(doc).find('a').each(() => {
+        const href = $(this).attr('href');
+
+        if (typeof href !== 'undefined' && !href.match(/^http:\/\//)) {
+          $(this).attr('href', $.path.addSearchParams(href, {
```
emteknetnz, line 180: will this add a duplicate search/querystring param if SubsiteID already exists?

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+            SubsiteID: subsiteId
+          }));
+        }
+      });
+
+      // Inject the SubsiteID as a hidden input into all forms submitting to the local site.
+      $(doc).find('form').each(() => {
+        const action = $(this).attr('action');
+
+        if (typeof action !== 'undefined' && !action.match(/^http:\/\//)) {
```
emteknetnz: also needs to match on https

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+          $(this).append(`<input type=hidden name="SubsiteID" value="${subsiteId}" >`);
```
emteknetnz: type="hidden"

client/src/legacy/entwine/SubsitesTreeDropdownField.js
```diff
@@ -0,0 +1,33 @@
+/* global jQuery */
+jQuery.entwine('ss', ($) => {
+  /**
+   * Choose a subsite from which to select pages.
+   * Needs to clear tree dropdowns in case selection is changed.
+   */
+  $('select.subsitestreedropdownfield-chooser').entwine({
+    onchange() {
+      // TODO Data binding between two fields
```
emteknetnz, line 9: will this TODO get done?

client/src/legacy/entwine/SubsitesTreeDropdownField.js
```diff
@@ -0,0 +1,33 @@
...
+   * before asking for the tree.
+   */
+  $('.TreeDropdownField.SubsitesTreeDropdownField').entwine({
+    getAttributes() {
+      const fieldName = this.attr('id').replace('Form_EditForm_', '');
+      const subsiteID = $(`#Form_EditForm_${fieldName}_SubsiteID option:selected`).val();
+
+      const attributes = this._super();
+      attributes.data.urlTree += `?${fieldName}_SubsiteID=${subsiteID}`;
+      attributes.data.cacheKey = `${attributes.data.cacheKey.substring(0, 19)}_${subsiteID}`;
```
emteknetnz, line 29: why are we getting a .substring(0, 19) ?

client/src/lib/changeActiveSubsite.js
```diff
@@ -0,0 +1,20 @@
+import createEvent from 'legacy/createEvent';
+
+export default function changeActiveSubsite(toSubsiteID, toSubsiteName) {
+  const { localStorage } = window;
```
emteknetnz: can get rid of this line and just use window.localStorage below

client/src/components/SubsiteChangeAlert/tests/SubsiteChangeAlert-test.js
```diff
@@ -0,0 +1,44 @@
...
+          revertCallback={callbackFn}
+        />
+      );
+      alert.instance().handleRevert();
+      expect(callbackFn.mock.calls).toEqual([['1', 'one']]);
+    });
+  });
+  describe('getMessage', () => {
+    it('should show the old subsite name correctly', () => {
+      const alert = shallow(
```
emteknetnz: rename this since window.alert() is an existing function, so pretty confusing have a variable with the same name

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+        subsiteName: $(`[value="${subsiteID}"]`, this).text()
+      };
+      window.localStorage.setItem('subsiteInfo', JSON.stringify(subsiteInfo));
+    },
+    showReactiveNotice(newSubsiteName) {
+      const { document } = window;
+      // React business
+      const modalContainer = this.getModalNode() || document.createElement('div');
+      document.body.appendChild(modalContainer);
+      const ChangeAlert = loadComponent('SubsiteChangeAlert');
```
emteknetnz: should be `const SubsiteChangeAlert` so that the jsx below has the correct component name?

client/src/components/SubsiteChangeAlert/SubsiteChangeAlert.jsx
```diff
@@ -0,0 +1,59 @@
...
+import i18n from 'i18n';
+import createEvent from 'legacy/createEvent';
+
+class SubsiteChangeAlert extends Component {
+  constructor(props) {
+    super(props);
+    this.handleRevert = this.handleRevert.bind(this);
+  }
+
+  handleRevert() {
```
emteknetnz: I'm finding the use of the word 'revert' throughout this confusing.  Shouldn't this be 'change' instead? (or click?)

client/src/legacy/createEvent.js
```diff
@@ -0,0 +1,20 @@
+/* global window */
+// Polyfill IE11
+function createEvent(type, extraData) {
+  const { document, Event } = window;
```
emteknetnz: what does this line do?  document and Event are already globally available

client/src/legacy/createEvent.js
```diff
@@ -0,0 +1,20 @@
...
+  let event;
+  if (typeof Event === 'object') {
+    event = document.createEvent('Event', true, true);
+    event.initEvent(type);
+  } else {
+    event = new Event(type);
+  }
+  if (extraData) {
+    Object.keys(extraData).forEach((key) => {
```
emteknetnz, line 18: `forEach(key =>`

client/src/legacy/createEvent.js
```diff
@@ -0,0 +1,20 @@
+/* global window */
+// Polyfill IE11
```
emteknetnz: this doesn't seem like it's a polyfill, might be better to remove this comment and put a comment in the `new Event(type)` block that this is for IE11

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+import { loadComponent } from 'lib/Injector';
+import createEvent from 'legacy/createEvent';
+import changeActiveSubsite from 'lib/changeActiveSubsite';
+
+jQuery.entwine('ss', ($) => {
+  $('#SubsitesSelect').entwine({
+    ModalNode: null,
+    onadd() {
+      // Storage has updated - another tab has changed the subsite info
+      window.addEventListener('storage', (storageEvent) => {
```
emteknetnz: `storage` is possibly a bit too generic, `subsitestorage` is more likely to never have a conflict

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
+/* jslint browser: true, nomen: true */
+/* global window, jQuery */
+import React from 'react';
+import ReactDom from 'react-dom';
+import { loadComponent } from 'lib/Injector';
+import createEvent from 'legacy/createEvent';
+import changeActiveSubsite from 'lib/changeActiveSubsite';
+
+jQuery.entwine('ss', ($) => {
```
emteknetnz: `$ =>` instead of `($) =>` - there's multiple instances where this can be updated, though I'll just mention it once here

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+    storeSubsiteInfo() {
+      const subsiteID = this.val();
+      const subsiteInfo = {
+        subsiteID,
+        subsiteName: $(`[value="${subsiteID}"]`, this).text()
+      };
+      window.localStorage.setItem('subsiteInfo', JSON.stringify(subsiteInfo));
+    },
+    showReactiveNotice(newSubsiteName) {
+      const { document } = window;
```
emteknetnz: don't need this line, document is already available

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+
+      if (!subsiteId) {
+        return;
+      }
+
+      // Inject the SubsiteID into internal links.
+      $(doc).find('a').each(() => {
+        const href = $(this).attr('href');
+
+        if (typeof href !== 'undefined' && !href.match(/^http:\/\//)) {
```
emteknetnz: need to match on https as well

client/src/legacy/entwine/LeftAndMain_Subsites.js
```diff
@@ -0,0 +1,190 @@
...
+  $('.cms-preview').entwine({
+
+    /**
+     * Update links and forms with GET/POST SubsiteID param, so we remaing on the current subsite.
+     * The initial link for the iframe comes from SiteTreeSubsites::updatePreviewLink.
+     *
+     * This is done so we can use the CMS domain for displaying previews so we prevent single-origin
+     * violations and SSL cert problems that come up when iframing content from a different URL.
+     */
+    onafterIframeAdjustedForPreview(event, doc) {
```
emteknetnz, line 168: where is this called from? there's no other reference to this function in this PR

client/src/lib/changeActiveSubsite.js
```diff
@@ -0,0 +1,20 @@
...
+  const request = new XMLHttpRequest();
+  const subsiteInfo = {
+    subsiteID: toSubsiteID,
+    subsiteName: toSubsiteName
+  };
+  request.open('GET', `?SubsiteID=${toSubsiteID}`);
```
emteknetnz: what is this actually doing? will this "refresh the whole CMS"?
is this something that's already done here? https://github.com/silverstripe/silverstripe-subsites/pull/443/files#diff-cf163c29c1a03e751d8439e2f36b5206R33

package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "silverstripe-subsites",
+  "description": "Builds the javascript for silverstripe subsites extension",
+  "main": "./client/src/boot/index.js",
+  "author": "You",
+  "engines": {
+    "node": ">=10.x"
```
emteknetnz: probably want to lock this to ^10 like other core packages e.g. https://github.com/silverstripe/silverstripe-asset-admin/blob/1/package.json#L17

client/src/styles/LeftAndMain_Subsites.css
```diff
@@ -0,0 +1,82 @@
...
+*/
+#SubsiteActions {
+  position: absolute;
+  padding: 0;
+  margin: 0;
+  right: 0;
+  top: 0;
+  width: 350px;
+  text-align: right;
+  margin-right: 130px;
```
emteknetnz, line 12: margin: 0 130px 0 0;

client/src/styles/LeftAndMain_Subsites.css
```diff
@@ -0,0 +1,82 @@
...
+  margin-left: 100px;
+}
+
+#Form_AddSubsiteForm label.left {
+  float: left;
+  width: 100px;
+  margin-left: -100px;
+}
+
+body.SubsiteAdmin .right form #URL .fieldgroup * {
```
emteknetnz, line 67: this selector looks a bit interesting, would it be possible to put a class="myclass" on something instead?

package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "scripts": {
+    "build": "yarn && yarn lint && yarn test && NODE_ENV=production webpack -p --bail --progress",
+    "dev": "NODE_ENV=development webpack --progress",
+    "watch": "NODE_ENV=development webpack --watch --progress",
+    "css": "WEBPACK_CHILD=css npm run build",
+    "test": "jest",
+    "coverage": "jest --coverage",
+    "lock": "npm-shrinkwrap --dev",
```
emteknetnz: shouldn't need shrinkwrap for anything since we are using yarn?

client/src/styles/LeftAndMain_Subsites.css
```diff
@@ -0,0 +1,82 @@
...
+  font-size: 11px;
+}
+
+.cms-add-form #PageType li .class-SubsitesVirtualPage {
+  background-position: 0 -32px !important;
```
emteknetnz, line 72: would it be possible to do this without using !important?

brynwhyman:
Hey @Cheddam you made a mention that you've been working on this recently. Do you have some local changes you could push up?

Cheddam:
@brynwhyman They're not quite ready to push, but I'll try and make time to get them polished off tomorrow.

client/src/legacy/entwine/LeftAndMain_Subsites.js
Mossman1215: Yeah that's the actual api name of the event type we're using.
https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent
https://developer.mozilla.org/en-US/docs/Web/API/Storage

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: This is not the Silverstripe way. All supported modules use .js for react components as well as other types of JS.
I actually agree, but consistency is a nice thing.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: No, this was existing code, moved in order to facilitate the new yarn build chain to support React components.
https://github.com/silverstripe/silverstripe-subsites/blob/fd10b868ec5dfccd8bd0273c92a30892e2af48cb/javascript/LeftAndMain_Subsites.js#L66 (current `2` branch HEAD)

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: Existing code outside scope of work for this PR
https://github.com/silverstripe/silverstripe-subsites/blob/fd10b868ec5dfccd8bd0273c92a30892e2af48cb/javascript/LeftAndMain_Subsites.js#L120

But no, it updates the param if it exists.

image

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: This is existing code, but seems like a bug we can easily fix. OK.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: Existing code, but we did refactor it to a backtick string without noticing that, good catch!

client/src/legacy/entwine/SubsitesTreeDropdownField.js
NightJar: Not in this PR.
Existing code, current head of `2` branch:
https://github.com/silverstripe/silverstripe-subsites/blob/fd10b868ec5dfccd8bd0273c92a30892e2af48cb/javascript/SubsitesTreeDropdownField.js#L9

client/src/legacy/entwine/SubsitesTreeDropdownField.js
NightJar: Because Entwine 🤷‍♂ 
Existing code, not altered in this PR - just moved into a new folder structure.
https://github.com/silverstripe/silverstripe-subsites/blob/fd10b868ec5dfccd8bd0273c92a30892e2af48cb/javascript/SubsitesTreeDropdownField.js#L27

client/src/legacy/createEvent.js
NightJar: Makes it very clear where these come from.
It's a small file so could they can be shifted into the global declaration atop the file though I suppose.

client/src/components/SubsiteChangeAlert/SubsiteChangeAlert.jsx
NightJar: No, revert describes what the intent is, not what the event is.
Can you think of a better suggestion that describes a users intent? We are struggling - longer forms still using "revert" in a more descriptive manner are all that's coming up.

client/src/legacy/createEvent.js
NightJar: 🤔 Interesting, I'm sure we did this because the linter was complaining, however it doesn't seem to be the case. It does bring a question of code style and consistency into play though...
E.g. I think the inclusion of parentheses makes `someCall('firstParam', firstParam => {...})` a little more readable.

client/src/legacy/createEvent.js
NightJar: Changed comment to state that it is a helper function instead.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: Indeed, this is not determined by us. It must be 'storage'.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: As with https://github.com/silverstripe/silverstripe-subsites/pull/443#discussion_r384174797 I'll not make a change until this discussion reaches an outcome.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: Seems like a duplicate comment for some reason 🤔 https://github.com/silverstripe/silverstripe-subsites/pull/443#pullrequestreview-349089514

NightJar: Oh, not they're different lines 😨 
Done.

client/src/legacy/entwine/LeftAndMain_Subsites.js
NightJar: Entwine.
I imagine it's not even in this module.

client/src/lib/changeActiveSubsite.js
NightJar: No, it won't update the CMS in terms of what the user sees, but it does "update the CMS" in regards of what the server keeps in session.

The line you referenced (`window.location.search = $.query.set('SubsiteID', subsiteID);`) does reload the CMS in the browser - this by contrast sends an XHR that updates the server and only fires update events on completion, meaning CMS users do not lose their draft.

This file should have this info in a docblock. We'll add one.

package.json
NightJar: Probably correct, although this was originally copied from another supported module, things change over time :)

client/src/styles/LeftAndMain_Subsites.css
NightJar: No, as this would override any other margin applied to those other dimensions. This specifically sets _only_ the right margin.

Also not a part of this PR. File is simply moved from original location.
https://github.com/silverstripe/silverstripe-subsites/blob/2/css/LeftAndMain_Subsites.css#L12

client/src/styles/LeftAndMain_Subsites.css
NightJar: Also not a part of this PR. File is simply moved from original location.
https://github.com/silverstripe/silverstripe-subsites/blob/2/css/LeftAndMain_Subsites.css#L67

I assume this has to do with specificity interplaying with other CMS UI rules.

package.json
NightJar: Probably correct. I'll remove it and see what happens.
Most amusingly it appears that shrinkwrap isn't listed as a dependency anyway 😂

client/src/styles/LeftAndMain_Subsites.css
NightJar: Also not a part of this PR. File is simply moved from original location.
https://github.com/silverstripe/silverstripe-subsites/blob/2/css/LeftAndMain_Subsites.css#L72

Cheddam:
I've resolved (as in clicked 'resolve' on) all of the feedback items, as they were either addressed or refuted. I've also pushed some additional tidy-up and tweaks that I made while reviewing this (only took me 2 months to get them up, sorry team 😅).

Final steps:

- [ ] Squash to one commit, ensuring all contributors are listed
- [ ] Rebase on `2` branch (this isn't a breaking change, so not sure why it's pointed at `master` - any thoughts @NightJar?)
- [ ] Merge and release

client/src/components/SubsiteChangeAlert/SubsiteChangeAlert.jsx
```diff
@@ -0,0 +1,79 @@
+/* global window */
```
emteknetnz: This file extension needs to be changed to `.js` to match the rest of the codebase

client/src/components/SubsiteChangeAlert/SubsiteChangeAlert.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Gets a translated string to display to the user
+   */
+  getMessage() {
+    const { newSubsiteName, currentSubsiteName } = this.props;
+
+    return i18n.inject(
+      i18n._t(
+        'SubsiteChangeAlert.SUBSITE_CHANGED',
```
emteknetnz: should this be added to `en.yml`?

client/src/components/SubsiteChangeAlert/SubsiteChangeAlert.jsx
```diff
@@ -0,0 +1,79 @@
...
+        `You have selected subsite "{newSubsiteName}" in another browser tab. In order to continue editing this content,
+        you must change the active subsite back to "{currentSubsiteName}".[SPLIT]Alternatively, you can reload this tab to switch to
+        subsite "{newSubsiteName}". Any unsaved changes you have made to this content will be lost.`
+      ),
+      {
+        newSubsiteName,
+        currentSubsiteName
+      }
+    ).split('[SPLIT]');
```
emteknetnz: It would be cleaner to return an array with two translatable strings `SUBSITE_CHANGED_LINE_ONE` and `SUBSITE_CHANGED_LINE_TWO`

client/src/legacy/entwine/SubsitesTreeDropdownField.js
emteknetnz, line 9: Please do something with this TODO, either remove it (preferable) or add some context to the comment e.g. linking it to something relevant so that future developers know what it is or when it can be safely removed

client/src/legacy/entwine/LeftAndMain_Subsites.js
emteknetnz: So, what you're saying is this TODO is just copy pasted from existing code

The link points to a PR merged in 2013, which is not really helpful

If so, just remove the TODO? It's currently just confusing

emteknetnz (changes requested):
It looks good, though a few small changes should be made

emteknetnz:
Behat failure is the preview/split mode in using the main site domain instead of the subsite domain.

Re-ran travis `2` https://travis-ci.org/github/silverstripe/silverstripe-subsites/builds/640118109 which is green

Have tried rebasing on top of the latest `2` locally and still getting same behat failure

Different issue - got this when running `yarn install & yarn build` (have already done the same for admin module)

image

emteknetnz:
Yeah, I'm not feeling this one as it is

I find the behat test failure to pretty concerning because it seems to show we've created a new layer in which to create werid bugs.

For instance, why do we need to do things like this? https://github.com/silverstripe/silverstripe-subsites/pull/443/files#diff-cf163c29c1a03e751d8439e2f36b5206R176

It seems as though we now need to hunt around and find all the things that now need to be updated via javascript just so that it works with this, new way of doing things?  I'm concerned that we're adding complexity and brittleness just to support a nice-to-have though not exactly essential feature.

Subsites already has a reputation for having weird bugs.  This PR will probably make things worse.

As it is I am not confident merging this. I'd like to see a decent chunk of this PR removed and simplified.  It may even need change of approach to get there.

Mossman1215:
> Yeah, I'm not feeling this one as it is
> 
> I find the behat test failure to pretty concerning because it seems to show we've created a new layer in which to create werid bugs.
> 
> For instance, why do we need to do things like this? https://github.com/silverstripe/silverstripe-subsites/pull/443/files#diff-cf163c29c1a03e751d8439e2f36b5206R176
> 
> It seems as though we now need to hunt around and find all the things that now need to be updated via javascript just so that it works with this, new way of doing things? I'm concerned that we're adding complexity and brittleness just to support a nice-to-have though not exactly essential feature.
> 
> Subsites already has a reputation for having weird bugs. This PR will probably make things worse.
> 
> As it is I am not confident merging this. I'd like to see a decent chunk of this PR removed and simplified. It may even need change of approach to get there.

It was supposed to be the simplest way possible to fix the issue where the user is editing in two tabs between subsites and changes their php session subsite id from what is in their tab cache.
it leads to issues where users are putting pages in the wrong subsite and raising tickets.

I think it's ok to close this PR and I'll think about a better approach.

emteknetnz:
@Mossman1215 I think the idea and general approach here is a good one and it's worth fixing.  The issue is that this particuarly javascript implementation is overly complicated and that's creates a maintenance problem on a module that's already complicated.

I'll close this one for now, though likely that we'll either reopen this in the future, or, start a new PR and reuse/reference the work done in this one

Issue #416, opened 2019-05-28
Subsites + Fluent = incorrect URL segment on edit


noizyboy:
I initially raised this over on the [SilverStripe forums,](https://forum.silverstripe.org/t/subsites-fluent-incorrect-url-segment-on-edit/1311) but haven't had any luck there, so giving it a go here. 

Version: Silverstripe CWP 2.1.1 (GitHub - silverstripe/cwp-installer: CWP project template)

When using Fluent in tandem with Subsites, the URLSegment shown on any subsite edit page uses the wrong base domain (making it difficult for the editor to open the page in another tab).

e.g. here’s the page without Fluent installed, showing the correct URL:

![de0d0cbc9198fec3b65cebeddc386a1c0b7f81cc](https://user-images.githubusercontent.com/416980/58446524-1ff71300-8155-11e9-92a6-4f324eaadae1.png)

…but if I add Fluent to my build, we get this…

![79fa97064e8cff2e45c541e8338368db390b47c9](https://user-images.githubusercontent.com/416980/58446521-1e2d4f80-8155-11e9-925d-91e5e1770bf8.png)

Any suggestions/ideas as to get the correct URL segment to show in that context?

robbieaverill:
Assuming "main.localhost" is NOT the configured domain for your "Subsite" subsite, and your screenshots are in the reverse order (the first is with fluent installed and the second is without), then I think there's a bug in fluent + subsites compatibility, probably in the DetectLocaleMiddleware, and we can investigate. Can you confirm my assumptions please?

noizyboy:
Ah, yes, sorry - uploaded images in correct order. (Fixed). 

And yes, it does appear to be directly related to fluent + subsites. Any insights welcome.

robbieaverill:
Is this affecting the URL segment preview in the CMS only? Or the frontend website as well?

noizyboy:
Just the CMS. Front-end is fine. 

(Although, having said that, we haven't added any translated content to the site yet, so the translated versions of those pages - and specifically the way the URL segments are generated - are currently an unknown quantity.)

robbieaverill:
Reproduced using CWP 2.3.x-dev (SilverStripe 4.4.x-dev with Subsites 2.3.x-dev and Fluent 4.2.x-dev).

robbieaverill:
Fixed in fluent with https://github.com/tractorcow-farm/silverstripe-fluent/pull/523

[...]

PR #444, opened 2019-10-11
FIX Subsites + Fluent = incorrect URL segment

mohamedallou:
fix for issue [Subsites + Fluent = incorrect URL segment on edit #416](https://github.com/silverstripe/silverstripe-subsites/issues/416)

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -2,6 +2,7 @@
 Name: subsiteconfig
 After:
   - 'framework/*'
+  - 'fluentcms'
 ---
 SilverStripe\AssetAdmin\Controller\AssetAdmin:
   treats_subsite_0_as_global: true


NightJar (approved):
Well the config change is safe, but I've not yet tested :)

emteknetnz:
What happens if fluent is not installed? Does having the `fluentcms` in the config file cause any issues?

dhensby:
> Does having the `fluentcms` in the config file cause any issues?

it should not

emteknetnz:
@Midouxx would you be able to retarget/rebase this pull-request to `2.3` instead of `master`.  Reasons being is that `master` is what would later become subsites `3.0`, while what we'd want to target is `2.3` which would be used for the next patch release of subsites

dhensby:
I've rebased this onto 2.3

emteknetnz:
Thanks Dan.  I'll give this a smoke test later this week then look to merge

emteknetnz:
@Midouxx 

This did not appear to fully work.  Possibly a different issue from the one originally linked on issue.  When I added a fluent locale, my subsite was no longer appeared in the url segment, e.g.

http://subsite.mydomain.test/home 

became

http://mydomain.test/home

emteknetnz (changes requested):
Need to have a better idea about issue above before this could be merged

emteknetnz:
This pull request hasn't had any activity for a while. Are you going to be doing further work on it, or would you prefer to close it now?

PR #447, opened 2019-11-27
Have JS work across different Form IDs

oilee80:
Default JS only works on `EditForm`
This was required for the module to work with linkable

### DIFF
--- a/javascript/SubsitesTreeDropdownField.js
+++ b/javascript/SubsitesTreeDropdownField.js
@@ -8,7 +8,8 @@
 			onchange: function() {
 				// TODO Data binding between two fields
                 const name = this.attr('name').replace('_SubsiteID', '');
-                let field = $('#Form_EditForm_' + name).first();
+                const formId = this.closest('form').attr('id');
+                let field = $('#' + formId + '_' + name).first();
                 field.setValue(0);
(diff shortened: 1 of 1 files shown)


javascript/SubsitesTreeDropdownField.js
```diff
@@ -8,7 +8,8 @@
 			onchange: function() {
 				// TODO Data binding between two fields
                 const name = this.attr('name').replace('_SubsiteID', '');
-                let field = $('#Form_EditForm_' + name).first();
+                const formId = this.closest('form').attr('id');
+                let field = $('#' + formId + '_' + name).first();
```
NightJar: I think that `const` is OK here, as the identifier `field` is never reassigned :)
as I understand it, mutation is OK, reassignment is not.
e.g. `const arr = []; arr.push('anything');` is fine.

NightJar (commented):
One super minor adjustment would be nice. But I shall not let it block a merge :)

NightJar:
Actually, that comment is pretty insignificant. It can easily be addressed at a later time. Thanks for your contribution! @oilee80 
image

oilee80:
Thanks @NightJar 😃

PR #451, opened 2020-01-09
Fix "Column 'URLSegment' in where clause is ambiguous" when duplicating pages

### LINKED ISSUE #450: Cannot duplicate page "Column 'URLSegment' in where clause is ambiguous"
When trying to duplicate certain pages, this SiteTreeSubsites extension throws the error  `Column 'URLSegment' in where clause is ambiguous`

While unable to identify the exact cause, as this only occurs on "some" pages, I assume its something to do with Fluent as well.

Converting `DataObject::get_one` to `SiteTree::get()->filter` seems to fix it

cjsewell:
Fixes #450

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -484,14 +484,10 @@ class SiteTreeSubsites extends DataExtension
 
                         $origDisableSubsiteFilter = Subsite::$disable_subsite_filter;
                         Subsite::disable_subsite_filter(true);
-                        $candidatePage = DataObject::get_one(
-                            SiteTree::class,
-                            "\"URLSegment\" = '"
-                            . Convert::raw2sql(urldecode($rest))
-                            . "' AND \"SubsiteID\" = "
-                            . $subsiteID,
-                            false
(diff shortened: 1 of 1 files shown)


src/Extensions/SiteTreeSubsites.php
```diff
@@ -484,14 +484,7 @@ public function augmentSyncLinkTracking()
...
                         Subsite::disable_subsite_filter(true);
-                        $candidatePage = DataObject::get_one(
-                            SiteTree::class,
-                            "\"URLSegment\" = '"
-                            . Convert::raw2sql(urldecode($rest))
-                            . "' AND \"SubsiteID\" = "
-                            . $subsiteID,
-                            false
-                        );
+                        $candidatePage = SiteTree::get()->filter(['URLSegment' => urldecode($rest), 'SubsiteID' => $subsiteID])->first();
```
NightJar: This line needs to be split in order to make the linter happy :)
I suggest making the array multi-line.

cjsewell: Fixed now

PR #452, opened 2020-01-27
Adjusting query used in getSubsiteIDForDomain

satrun77:
The same issue as #287. Fixed in version 1.x but not in 2.x


emteknetnz (changes requested):
@satrun77 looks like there's a bunch of changes from other pull-requests.  You'll need to rebase/retarget this PR to the `2.3` branch so that we can release this as a patch.  Right now this is targetting `master` which would be used to release `3.0`

PR #458, opened 2020-11-05
Fix yaml config, use sminnee/phpunit fork

dnsl48:


.travis.yml
```diff
@@ -17,17 +17,17 @@ env:
 matrix:
   include:
     - php: 5.6
-      env: DB=MYSQL INSTALLER_VERSION=4.4.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
+      env: DB=MYSQL INSTALLER_VERSION=~4.4.0@dev PHPCS_TEST=1 PHPUNIT_TEST=1
     - php: 7.0
```
emteknetnz: Remove php 5.6 + 7.0 from matrix

Add php 7.4 + 8.0 (nightly)

Should only have a single `4.4.0@dev` entry in the matrix for php 7.1

Should have 4.5.x-dev, 4.6.x-dev (maybe 4.7.x-dev) and 4.x-dev

emteknetnz (changes requested):

.travis.yml
dnsl48: then we'll have to upgrade to xenial and maybe then it's actually easier to go for shared config...

dnsl48: done

.travis.yml
```diff
@@ -1,64 +1,10 @@
-language: php
+# https://docs.travis-ci.com/user/build-config-imports/
```
emteknetnz: May as well remove this comment

emteknetnz (approved):
MOG

PR #460, opened 2021-03-16
Swap column with value in the where condition when filtering for matching domains

michalkleiner:
Not 100% sure why the comparison was the original (other-way round) way, but it didn't appear to work.


michalkleiner:
I see now, needs a bit more work then.

PR #473, opened 2021-12-06
ENH Add domain support for the main site.

### LINKED ISSUE #472: Main site should have a domain
The main site should have a domain associated with it.

The specific use case I have right now is for absolute URLs in emails - subsites have domains associated with them so calls to `AbsoluteLink` result in the appropriate URL for any pages belonging to a subsite - but for main site absolute URLs (especially when using a queued job via the cli where there is no host to draw from) I have to work around it by storing the domain somewhere effectively arbitrary.

If the main site had a domain the same way that subsites do, getting an absolute URL for the main site would be straight forward.

(for the absolute URL to automatically apply correctly in an email, we'd need to make a change to `Director::absoluteURL` (see silverstripe/silverstripe-framework#10168) - that is obviously out of scope for this issue, which is dealing with getting a correct absolute URL when using `AbsoluteLink` on a page which I am currently doing to set absolute URLs prior to setting the email body)

GuySartorelli:
Fixes #472 

This ensures the `AbsoluteLink` for pages will always be correct for main site pages (if a domain is set up for the main site) - even when calling it from the context of a subsite. This allows easily linking to main site pages from subsites, as well as any other situation where the absolute URL for the main site needs to be fetched correctly.

### DIFF
--- a/src/Admin/SubsiteAdmin.php
+++ b/src/Admin/SubsiteAdmin.php
@@ -7,6 +7,7 @@ use SilverStripe\Forms\GridField\GridFieldDetailForm;
 use SilverStripe\Forms\GridField\GridFieldPaginator;
 use SilverStripe\Subsites\Forms\GridFieldSubsiteDetailForm;
 use SilverStripe\Subsites\Model\Subsite;
+use SilverStripe\Subsites\Model\SubsiteDomain;
 
 /**
  * Admin interface to manage and create {@link Subsite} instances.
@@ -15,7 +16,14 @@ use SilverStripe\Subsites\Model\Subsite;
  */
 class SubsiteAdmin extends ModelAdmin
 {
-    private static $managed_models = [Subsite::class];
+    private static $managed_models = [
+        Subsite::class => [
+            'title' => 'Subsites'
+        ],
+        SubsiteDomain::class => [
+            'title' => 'Main site domains'
+        ],
+    ];
 
     private static $url_segment = 'subsites';
 
@@ -31,11 +39,18 @@ class SubsiteAdmin extends ModelAdmin
     {
         $form = parent::getEditForm($id, $fields);
 
-        $grid = $form->Fields()->dataFieldByName(str_replace('\\', '-', Subsite::class));
-        if ($grid) {
-            $grid->getConfig()->getComponentByType(GridFieldPaginator::class)->setItemsPerPage(100);
-            $grid->getConfig()->removeComponentsByType(GridFieldDetailForm::class);
-            $grid->getConfig()->addComponent(new GridFieldSubsiteDetailForm());
+        if ($this->modelClass === Subsite::class) {
+            $grid = $form->Fields()->dataFieldByName(str_replace('\\', '-', Subsite::class));
+            if ($grid) {
+                $grid->getConfig()->getComponentByType(GridFieldPaginator::class)->setItemsPerPage(100);
+                $grid->getConfig()->removeComponentsByType(GridFieldDetailForm::class);
+                $grid->getConfig()->addComponent(new GridFieldSubsiteDetailForm());
+            }
+        }
+
+        if ($this->modelClass === SubsiteDomain::class) {
+            $grid = $form->Fields()->dataFieldByName($this->sanitiseClassName(SubsiteDomain::class));
+            $grid->setList($grid->getList()->filter('SubsiteID', 0));
         }
 
         return $form;
--- a/src/Admin/SubsiteAdmin.php
+++ b/src/Admin/SubsiteAdmin.php
@@ -40,7 +40,7 @@ class SubsiteAdmin extends ModelAdmin
         $form = parent::getEditForm($id, $fields);
 
         if ($this->modelClass === Subsite::class) {
-            $grid = $form->Fields()->dataFieldByName(str_replace('\\', '-', Subsite::class));
+            $grid = $form->Fields()->dataFieldByName($this->sanitiseClassName(Subsite::class));
             if ($grid) {
                 $grid->getConfig()->getComponentByType(GridFieldPaginator::class)->setItemsPerPage(100);
                 $grid->getConfig()->removeComponentsByType(GridFieldDetailForm::class);
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -412,9 +412,7 @@ class SiteTreeSubsites extends DataExtension
         // Generate the existing absolute URL and replace the domain with the subsite domain.
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
-        if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url);
-        }
+        $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url);
         return $url;
     }
 
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -885,6 +885,16 @@ class Subsite extends DataObject
      */
     public function getPrimarySubsiteDomain()
     {
+        // Main site
+        if (!$this->isInDB()) {
+            Subsite::disable_subsite_filter(true);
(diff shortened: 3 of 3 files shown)


GuySartorelli:
Note: In managed models, the `title` was set explicitly for the subsites tab to avoid the main site domains tab becoming the first tab.

src/Admin/SubsiteAdmin.php
```diff
@@ -31,11 +39,18 @@ public function getEditForm($id = null, $fields = null)
...
+            $grid = $form->Fields()->dataFieldByName(str_replace('\\', '-', Subsite::class));
+            if ($grid) {
+                $grid->getConfig()->getComponentByType(GridFieldPaginator::class)->setItemsPerPage(100);
+                $grid->getConfig()->removeComponentsByType(GridFieldDetailForm::class);
+                $grid->getConfig()->addComponent(new GridFieldSubsiteDetailForm());
+            }
+        }
+
+        if ($this->modelClass === SubsiteDomain::class) {
+            $grid = $form->Fields()->dataFieldByName($this->sanitiseClassName(SubsiteDomain::class));
```
michalkleiner, line 52: Would this `sanitiseClassName` call work for the Subsite::class branch above as well to keep it consistent?

src/Model/Subsite.php
```diff
@@ -885,6 +885,16 @@ public function domain()
      */
     public function getPrimarySubsiteDomain()
     {
+        // Main site.
```
michalkleiner:
```suggestion
        // Main site
```

src/Model/Subsite.php
```diff
@@ -885,6 +885,16 @@ public function domain()
...
+        if (!$this->isInDB()) {
+            Subsite::disable_subsite_filter(true);
+            $domain = SubsiteDomain::get()->filter('SubsiteID', 0)
+                ->sort('"IsPrimary" DESC')
+                ->first();
+            Subsite::disable_subsite_filter(false);
+            return $domain;
+        }
+        // Subsites.
```
michalkleiner:
```suggestion
        // Subsites
```

src/Admin/SubsiteAdmin.php
```diff
@@ -15,7 +16,14 @@
  */
 class SubsiteAdmin extends ModelAdmin
 {
-    private static $managed_models = [Subsite::class];
+    private static $managed_models = [
+        Subsite::class => [
+            'title' => 'Subsites'
+        ],
+        SubsiteDomain::class => [
+            'title' => 'Main Site Domains'
```
michalkleiner:
```suggestion
            'title' => 'Main site domains'
```

michalkleiner (commented):
Thanks @GuySartorelli, the approach looks sensible. I've added a few minor inline comments if you could look at those, thanks!

src/Admin/SubsiteAdmin.php
ChloeSartorelli: Yup. I was considering doing that but I wasn't sure if I should do it in this pull request since it's _technically_ not part of the same change. I'll make that change along with the other minor suggestions you've made :)

michalkleiner:
Though I always thought the main site has mechanisms to get its primary domain anyway, with or without using the subsites module. Is the issue rather in the bespoke code/additions or modules where it doesn't correctly pick the subsite e.g. from CLI or similar, and configs like `alternate_base_url` should be used instead?

GuySartorelli:
> Thanks @GuySartorelli, the approach looks sensible. I've added a few minor inline comments if you could look at those, thanks!

Sweet! I'll make those changes shortly.

> Though I always thought the main site has mechanisms to get its primary domain anyway, with or without using the subsites module. Is the issue rather in the bespoke code/additions or modules where it doesn't correctly pick the subsite e.g. from CLI or similar, and configs like `alternate_base_url` should be used instead?

So... in some situations the main site has mechanisms to get its primary domain. And sometimes they work. Sometimes they don't. For the times it doesn't work I have added a PR to allow that to be configured dynamically (see silverstripe/silverstripe-framework#10168) but that's a bit out of scope.

`alternate_base_url` is fine for most situations, but I'm often in a situation where the client hasn't made a final decision on the production domain until after we've already deployed code to the production environment, so I always end up adding in a field to the `SiteConfig` to store the main site domain anyway. This PR gives a more consistent place to set that domain and means I don't have to add it myself every time I add subsites to a codebase.

GuySartorelli:
@michalkleiner I've made the requested changes.

src/Model/Subsite.php
```diff
@@ -885,6 +885,16 @@ public function domain()
      */
     public function getPrimarySubsiteDomain()
     {
+        // Main site
+        if (!$this->isInDB()) {
```
michalkleiner, line 889: Now seeing the method this is in and the not-in-db check... is there a better/different place this could go to? Alternatively, can we get more explanation why it's here and why we need to check it's not in the db?

ChloeSartorelli: An alternative would be `if (!$this->ID) {`.
This is for situations like `SiteTreeSubsites::alternateAbsoluteLink` (inline below for convenience) where we call `domain` (which calls `getPrimarySubsiteDomain`) on the relationship directly from the `has_one` relation object.
If the page has no `SubiteID` (i.e. it belongs to the main site), the relation object will still be an instantiated `Subsite` object which has no ID and which returns `false` from `isInDB` - so we end up getting the main site domain.

I can't think of somewhere else to put the call, since the intention is to distinguish between a main site or subsite `Subsite` object (where a main site `Subsite` object is any `Subsite` object without an ID, effectively) when fetching the domain.

This allows project-specific code or code from other modules to remain the same (if they don't have the "if there is a subsite ID" check in place - in that case there's still only a very minor change required) and now get the added benefit of automagically fetching the correct main site domain as well.

https://github.com/silverstripe/silverstripe-subsites/blob/0b7a63e0482a8e004ef34595233654124189fcbe/src/Extensions/SiteTreeSubsites.php#L410-L417

ChloeSartorelli: Just realised I didn't explicitly say why we need to check for the main site vs an actual subsite there - the main site domains aren't stored on the Domains relation of any subsite object, so we can't just use `$this->Domains()` for the main site.

src/Extensions/SiteTreeSubsites.php
```diff
@@ -412,9 +412,7 @@ public function alternateAbsoluteLink($action = null)
         // Generate the existing absolute URL and replace the domain with the subsite domain.
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
-        if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url);
-        }
+        $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url);
```
emteknetnz, line 415: Why are we removing this check? If $this->owner->SubsiteID == 0 then this will cause an exception i.e. `(null)->domain()`

emteknetnz (changes requested):
I'm worried this PR has a high chance of causing regressions when people run `composer update`

What will the experience be like for all existing sites that do not have SubsiteDomain's defined once this code is released?

Is this 100% backwards compatible with "Incomplete data"? Or is there an expectation that everyone will fill create a new SubsiteDomain record (they won't)

I'd need unit tests testing things like `AbsoluteLink()` on this PR both with and without a SubsiteDomain record on the main site to feel comfortable here

GuySartorelli:
Honestly it's been far too long since I've touched this PR. It needs work and I don't have the time to do it and we've found ways around this in the project I was working on at the time. Closing, but someone else can pick it up if they need something like this in their projects and haven't found their own workaround.

src/Extensions/SiteTreeSubsites.php
ChloeSartorelli: I'm about 60% sure `$this->owner->Subsite()` would give a new Subsite object with no ID - but that would need to be checked if someone picks this PR up again. If that's not the case, then another mechanism will need to be added here to get the main site Domain where no subsite is set.

NightJar: I'm 100% sure.
https://github.com/silverstripe/silverstripe-framework/blob/6d90eec0c883a69a87e0ad31d0e8ba617b809057/src/ORM/DataObject.php#L1890

PR #476, opened 2022-02-15
FIX replace in_array check with `hasTable` check.

wilr:
in_array is case-sensitive and may not detect that a table exists when using lowercase format


michalkleiner (approved):

wilr:
thanks @michalkleiner

PR #481, opened 2022-06-21
added config setting to ignore subsite language; when using subsites …

hubertusanton:
subsite in combination with fluent it picks up wrong yml file because this method sets subsites language.

SilverStripe\Subsites\Extensions\SiteTreeSubsites:
  ignore_subsite_language: true


GuySartorelli:
@hubertusanton what version of fluent are you using?

@mfendeksilverstripe I know you've done a lot of work with fluent - have you encountered this problem before?

I've just a bit weary of merging a change like this if it's something that fluent is already, or potentially should take into account.

mfendeksilverstripe:
@GuySartorelli I suggest to create a `LocaleDetector` for subsites. Fluent module provides an interface [LocaleDetector](https://github.com/tractorcow-farm/silverstripe-fluent/blob/5/src/State/LocaleDetector.php) to allow high degree of extensibility. Alternatively, you can subclass and override the default [BrowserLocaleDetector](https://github.com/tractorcow-farm/silverstripe-fluent/blob/5/src/State/BrowserLocaleDetector.php).

There are [docs available](https://github.com/tractorcow-farm/silverstripe-fluent/blob/5/docs/en/locale-detection.md) covering this topic.

The goal would be to provide a locale detection system for subsites so we keep the code nice and clean :)

hubertusanton:
@GuySartorelli I'm using silverstripe/subsites 2.6.0

hubertusanton:
The problem is that subsites sets i18n based on the locale setting of the current subsite.
When not doing when using the combi fluent/ subsites it's fixed. I'll change the pull request, it's not a config setting anymore.

GuySartorelli:
@hubertusanton Given that this isn't something Fluent takes into consideration automatically, and it doesn't sound like there are plans for it to do so, I think this should be a config option like you had originally. Subsites shouldn't need to know what the Fluent module is - and anyone using some other module for localisation shouldn't be punished for not using Fluent by having this same problem, with no way around it.
Can you please change it back to a configuration option, and add some documentation?

src/Extensions/SiteTreeSubsites.php
```diff
@@ -398,8 +398,9 @@ public static function contentcontrollerInit($controller)
             SSViewer::set_themes(ThemeResolver::singleton()->getThemeList($subsite));
         }
 
-        if ($subsite && i18n::getData()->validate($subsite->Language)
-            && !singleton(SiteTree::class)->hasExtension('TractorCow\Fluent\Extension\FluentSiteTreeExtension')) {
+        $ignore_subsite_language = Config::inst()->get(self::class, 'ignore_subsite_language');
+
+        if ($subsite && i18n::getData()->validate($subsite->Language) && !$ignore_subsite_language) {
```
ChloeSartorelli:
```suggestion
        if (!$ignore_subsite_language && $subsite && i18n::getData()->validate($subsite->Language)) {
```
No need to call `validate()` if we're ignoring the locale anyway.

GuySartorelli (changes requested):
Can you please call the config `ignore_subsite_locale` instead? And for the PHP variable, use lowerCamelCase.
So `$ignoreSubsiteLocale = Config::inst()->get(self::class, 'ignore_subsite_locale');`

README.md
```diff
@@ -220,6 +220,14 @@ or by defining the subsiteCMSShowInMenu function in your admin:
 		return true;
 	}
 
+### Using Subsites in combination with Fluent
+
+When using Subsites in combination with Fluent module, the Subsites module sets the i18n language to the language defined in the current Subsite. When this behaviour is not desired and you need to use the language in FluentState use the following setting in your yml config file:
```
ChloeSartorelli: Locale, not language.

GuySartorelli:
I should probably ask for a unit test as well - can you please add one?

src/Extensions/SiteTreeSubsites.php
hubertusanton: ok, i'll change this.

hubertusanton:
Did some work on a test, can you check this if i'm doing this right?

In SiteTreeSubsitesTest.php created this method:

`    public function testIgnoreSubsiteLocale() {

        $ignore_subsite_locale = Config::inst()->set(SiteTreeSubsites::class, 'ignore_subsite_locale', true);

        $subsitePage = $this->objFromFixture(Page::class, 'subsite_locale_about');
        Subsite::changeSubsite($subsitePage->SubsiteID);
        $controller = ModelAsController::controller_for($subsitePage);

        ray('i18n locale before extend: ' . i18n::get_locale());
        SiteTree::singleton()->extend('contentcontrollerInit', $controller);
        ray('i18n locale after extend: ' . i18n::get_locale());

        $i18n_locale = i18n::get_locale();

        $this->assertEquals('en_US', $i18n_locale);

    }`

hubertusanton:
added the test, let me know if this is ok

GuySartorelli:
Some of the tests are failing - can you please take a look?

hubertusanton:
> Some of the tests are failing - can you please take a look?

done

GuySartorelli:
Awesome, thank you. However you've added a `.idea/workspace.xml` file while doing so - can you please remove that file?

hubertusanton:
> Awesome, thank you. However you've added a `.idea/workspace.xml` file while doing so - can you please remove that file?

nooby mistake srry its gone now :)

src/Extensions/SiteTreeSubsites.php
```diff
@@ -398,8 +398,12 @@ public static function contentcontrollerInit($controller)
             SSViewer::set_themes(ThemeResolver::singleton()->getThemeList($subsite));
         }
 
-        if ($subsite && i18n::getData()->validate($subsite->Language)) {
-            i18n::set_locale($subsite->Language);
+        $ignore_subsite_locale = Config::inst()->get(self::class, 'ignore_subsite_locale');
+
+        if ($subsite && $subsite->Language) {
```
ChloeSartorelli: Why was this added? We're already checking `$subsite` for truthiness if the below condition so all this adds is a check on truthiness for `$subsite->Language` - is there a chance this doesn't have a value?
If so, please just add it to the condition below. Nested conditions like this are a bit of an anti-pattern.

GuySartorelli (changes requested):

src/Extensions/SiteTreeSubsites.php
hubertusanton: Had the problem of $subsite existing and $subsite->Language had no value. When no subsite is defined in cms this throws a Notice. I moved this to the if below now.

src/Extensions/SiteTreeSubsites.php
```diff
@@ -400,11 +400,10 @@ public static function contentcontrollerInit($controller)
 
         $ignore_subsite_locale = Config::inst()->get(self::class, 'ignore_subsite_locale');
 
-        if ($subsite && $subsite->Language) {
-            if (!$ignore_subsite_locale && $subsite && i18n::getData()->validate($subsite->Language)) {
-                i18n::set_locale($subsite->Language);
-            }
+        if (!$ignore_subsite_locale && $subsite && i18n::getData()->validate($subsite->Language) && $subsite->Language) {
```
ChloeSartorelli:
```suggestion
        if (!$ignore_subsite_locale && $subsite && $subsite->Language && i18n::getData()->validate($subsite->Language)) {
```
If there's no language we shouldn't pass the non-existent language into the validate method.

GuySartorelli (changes requested):

GuySartorelli (approved):
Looks good and works locally. Thank you for doing this.

PR #490, opened 2022-09-27
Switch to listing views of subsite-filtered sections

### LINKED ISSUE #404: Switching subsites when a subsite-specific file/folder is open breaks the detail view
**Version: CWP 2.2.x-dev (subsites 2.2)**

* Set up two subsites
* Go into the assets area
* Switch to a non-default subsite
* Create a new file or folder
* Set it to belong to the secondary subsite and save
* With the detail screen for the file/folder still open, switch to another subsite

In some cases it would show me the detail view of the folder which should not be visible in that subsite, and in other cases it throws an error:

image

I'm not sure what the expected behaviour should be really. Perhaps there needs to be a change to asset-admin to handle this scenario (the folder doesn't exist in the current context, and/or the response to get its details fails).

### LINKED ISSUE #461: Cannot switch subsites in the CMS when on the 'Settings' section of a page
Subsites version: 2.4.0
CMS version: 4.7.0
Browser: Chrome 89.0

Steps to replicate:

1. Setup a SilverStripe instance with a main site and one or more subsites.
2. Create and publish a page on the main site.
3. Navigate back to the page within the CMS and change to the 'Settings' tab.
4. Switch to a subsite using the subsite dropdown in the CMS navigation bar.

Expected: the CMS should switch to the 'Pages' view of the selected subsite

Actual: the CMS reloads the page already being viewed and remains on the 'Main site' rather than the selected subsite.

This bug also appears to occur if you're on the 'History' tab.

### LINKED ISSUE #489: Subsite switching: open page appears to move along to new subsite
When switching to another subsite while having a specific page open in edit mode, that page will be re-opened in the other subsite as if it has move there. This is confusing and it seems there are also [some](https://github.com/silverstripe/silverstripe-subsites/issues/404) [other](https://github.com/silverstripe/silverstripe-subsites/issues/461) issues related to this behaviour. It seems the system just finds the page by its ID and opens it in edit mode (even though it should be filtered out because it belongs to another subsite).

I propose we remove the specific ID (of any currently edited file/folder/page/settings/history/etc) when switching subsites. That way you'd always start with an overview of all pages/files/etc in the subsite you've just switched to.

```JS
                    // Switch to listing views of subsite-filtered sections (eg: pages/ & assets/)
                    let newLocHref = window.location.href
                        .split('edit/show/').shift() // remove [pages/]edit/etc
                        .split('settings/show/').shift() // remove [pages/]settings/etc
                        .split('history/show/').shift() // remove [pages/]history/etc
                        .split('show/').shift(); // remove [assets/]show/102/edit/etc (last as 'show' is broad)
                    window.location.href = newLocHref + $.query.set('SubsiteID', $(this).val()).toString();
```
PR coming up.

micschk:
Fixes #489
Fixes #404 
Fixes #461

### DIFF
--- a/client/javascript/LeftAndMain_Subsites.js
+++ b/client/javascript/LeftAndMain_Subsites.js
@@ -7,7 +7,13 @@
 		$('#SubsitesSelect').entwine({
 			onadd:function(){
 				this.on('change', function(){
-					window.location.search=$.query.set('SubsiteID', $(this).val());
+                    // Switch to listing views of subsite-filtered sections (eg: pages/ & assets/)
+                    let newLocHref = window.location.href
+                        .split('edit/show/').shift() // remove [pages/]edit/etc
+                        .split('settings/show/').shift() // remove [pages/]settings/etc
+                        .split('history/show/').shift() // remove [pages/]history/etc
+                        .split('show/').shift(); // remove [assets/]show/102/edit/etc (last as 'show' is broad)
+                    window.location.href = newLocHref + $.query.set('SubsiteID', $(this).val()).toString();
 				});
 			}
 		});


GuySartorelli (approved):
This looks great and works well locally, thank you! I've retargetted this to the `2.8` branch  so we can release it as a patch, since it's fixing a bug.

emteknetnz:
This one looks to have caused a regression - I've opened a PR to revert https://github.com/silverstripe/silverstripe-subsites/pull/539

GuySartorelli:
@micschk Sorry we had to revert this PR - there's more information in https://github.com/silverstripe/silverstripe-subsites/issues/538. If you'd like to take another go at this I'd be happy to review it, just `@` me.

Issue #495, opened 2022-11-30
4.12 regression: "Copy to subsite" button shows in compare mode


maxime-rainville:
![image](https://user-images.githubusercontent.com/1168676/204721578-e1d2a9d6-6c2d-4bd4-a844-436aba20b365.png)

## Acceptance criteria
- [x] Copy to subsite button is not present in the compare view

## PRs
- https://github.com/silverstripe/silverstripe-subsites/pull/496

GuySartorelli:
PR merged

PR #496, opened 2022-12-09
FIX Do not show copy to subsite buttons in history viewer

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-subsites/issues/495

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -29,6 +29,7 @@ use SilverStripe\Subsites\Model\Subsite;
 use SilverStripe\Subsites\Service\ThemeResolver;
 use SilverStripe\Subsites\State\SubsiteState;
 use SilverStripe\View\SSViewer;
+use SilverStripe\VersionedAdmin\Controllers\HistoryViewerController;
 
 /**
  * Extension for the SiteTree object to add subsites support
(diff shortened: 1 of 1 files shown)


src/Extensions/SiteTreeSubsites.php
```diff
@@ -117,10 +118,21 @@ public function updateCMSFields(FieldList $fields)
...
+        // Check if we're viewing page history
+        // e.g. request to "/admin/historyviewer/schema/versionForm/2"
+        $viewingPageHistory = false;
+        if (class_exists(HistoryViewerController::class)) {
+            $urlSegment = HistoryViewerController::config()->get('url_segment');
+            $request = Controller::curr()->getRequest();
+            if (strpos($request->getURL(), "/$urlSegment/") !== false) {
+                $viewingPageHistory = true;
+            }
+        }
```
ChloeSartorelli:
```suggestion
        $viewingPageHistory = Controller::has_curr() && Controller::curr() instanceof HistoryViewerController;
```
I think this is a lot cleaner, and will catch unexpected edge cases (e.g. custom `HistoryViewerController` subclass with a different `url_segment`) or some custom route that uses `HistoryViewerController` without respecting `url_segment`.

GuySartorelli (changes requested):

src/Extensions/SiteTreeSubsites.php
emteknetnz: Ah yep much better

Have updated - note I got rid of the `Controller::has_curr()` as `var_dump(null instanceof stdClass);` is `false`

ChloeSartorelli: For anyone seeing this in the future: We spoke offline about that - Steve has added `Controller::has_curr()` back in, since `Controller::curr()` throws an error if there's no current controller.

GuySartorelli (approved):

PR #499, opened 2023-01-18
API Remove deprecated code

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10546

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\Subsites\Extensions;
 
-use SilverStripe\Dev\Deprecation;
 use Page;
 use SilverStripe\CMS\Forms\SiteTreeURLSegmentField;
 use SilverStripe\CMS\Model\SiteTree;
@@ -10,7 +9,6 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTP;
 use SilverStripe\Core\Config\Config;
-use SilverStripe\Core\Convert;
 use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
@@ -77,10 +75,15 @@ class SiteTreeSubsites extends DataExtension
         }
 
         $subsiteID = null;
-        if (Subsite::$force_subsite) {
-            $subsiteID = Subsite::$force_subsite;
+        $subsiteState = SubsiteState::singleton();
+        $force_subsite = $subsiteState->withState(function (SubsiteState $newState) {
+            return $newState->getSubsiteId();
+        });
+
+        if ($force_subsite) {
+            $subsiteID = $force_subsite;
         } else {
-            $subsiteID = SubsiteState::singleton()->getSubsiteId();
+            $subsiteID = $subsiteState->getSubsiteId();
         }
 
         if ($subsiteID === null) {
@@ -423,7 +426,11 @@ class SiteTreeSubsites extends DataExtension
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
         if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url ?? '');
+            $pattern = '/\/\/[^\/]*(?<slash>\/)?/';
+            preg_match($pattern, $url ?? '', $matches, PREG_UNMATCHED_AS_NULL);
+            $delim = $matches['slash'] ?? '';
+
+            $url = preg_replace($pattern, '//' . $this->owner->Subsite()->domain() . $delim, $url ?? '');
         }
         return $url;
     }
@@ -442,22 +449,6 @@ class SiteTreeSubsites extends DataExtension
         $url = Director::absoluteURL($this->owner->Link($action));
         $link = HTTP::setGetVar('SubsiteID', $this->owner->SubsiteID, $url);
         return $link;
-    }
-
-    /**
-     * This function is marked as deprecated for removal in 5.0.0 in silverstripe/cms
-     * so now simply passes execution to where the functionality exists for backwards compatiblity.
-     * CMS 4.0.0 SiteTree already throws a SilverStripe deprecation error before calling this function.
-     * @deprecated 2.2.0 Use updatePreviewLink() instead
-     *
-     * @param string|null $action
-     * @return string
-     */
-    public function alternatePreviewLink($action = null)
-    {
-        Deprecation::notice('2.2.0', 'Use updatePreviewLink() instead');
-        $link = '';
-        return $this->updatePreviewLink($link, $action);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Extensions/SiteTreeSubsites.php
```diff
@@ -423,7 +426,11 @@ public function alternateAbsoluteLink($action = null)
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
         if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url ?? '');
+            $pattern = '/\/\/[^\/]*(?<slash>\/)?/';
+            preg_match($pattern, $url ?? '', $matches, PREG_UNMATCHED_AS_NULL);
+            $delim = $matches['slash'] ?? '';
+
+            $url = preg_replace($pattern, '//' . $this->owner->Subsite()->domain() . $delim, $url ?? '');
```
emteknetnz, line 433:
```suggestion
            $url = preg_replace('#^(https?://)[^/]+#', '$1' . $this->owner->Subsite()->domain(), $url);
```

src/Extensions/SiteTreeSubsites.php
```diff
@@ -524,23 +515,26 @@ public function augmentSyncLinkTracking()
...
+        return SubsiteState::singleton()->withState(function (SubsiteState $newState) {
+            
+            // If this page is being filtered in the current subsite, then no custom validation query is required.
+            $subsite = $newState->getSubsiteId() ?: SubsiteState::singleton()->getSubsiteId();
+            if (empty($this->owner->SubsiteID) || $subsite == $this->owner->SubsiteID) {
+                return null;
+            }
+            
+            // Backup forced subsite
+            $prevForceSubsite = $newState->getSubsiteId();
```
emteknetnz, line 527: This seems wrong, we shouldn't need to repeat the previous logic backing up force_subsite inside the new withState() method, the point of the with*() pattern of methods is they reset themselves

src/Extensions/SiteTreeSubsites.php
```diff
@@ -524,23 +515,26 @@ public function augmentSyncLinkTracking()
...
-        Subsite::$force_subsite = $this->owner->SubsiteID;
-
-        // Repeat validation in the correct subsite
-        $isValid = $this->owner->validURLSegment();
-
-        // Restore
-        Subsite::$force_subsite = $prevForceSubsite;
-
-        return (bool)$isValid;
+        return SubsiteState::singleton()->withState(function (SubsiteState $newState) {
```
emteknetnz, line 518: This entire block seems wrong

Have a look at other implementations of withState() on extensions within the subsites module e.g. `FileSubsites::canEdit()`

```php
        return SubsiteState::singleton()->withState(function (SubsiteState $newState) use ($member) {
            $newState->setSubsiteId($this->owner->SubsiteID);

            return $this->owner->getPermissionChecker()->canEdit($this->owner->ID, $member);
        });
```

You're supposed to call `$newState->setSubsiteId($subsiteID);` on the first line of the callback

src/Reports/SubsiteReportWrapper.php
```diff
@@ -64,13 +65,17 @@ public function beforeQuery($params)
     {
         // The user has select a few specific sites
         if (!empty($params['Subsites'])) {
-            Subsite::$force_subsite = $params['Subsites'];
+            SubsiteState::singleton()->withState(function (SubsiteState $newState) use ($params) {
```
emteknetnz, line 68: The previous logic with presumably setting the subsite to use for the rest of the request?

This block is doing absolutely nothing?  Calling `$newState->setSubsiteID()` within `withState` set the subsiteID for the duration of the block, not the rest of the request

The comment also applies to the withState() calls below in this PR

emteknetnz (changes requested):

src/Reports/SubsiteReportWrapper.php
sabina-talipova: I also think that these two methods do nothing. Should we mark them as deprecated and replace with new methods that we invoke callback function.
```php
    public function beforeAfterQuery($callback, $args, $params)
    {
        return SubsiteState::singleton()->withState(function (SubsiteState $newState) use ($callback, $args, $params) {

            if (!empty($params['Subsites'])) {
                $newState->setSubsiteId($params['Subsites']);
            } else {
                $subsites = Subsite::accessible_sites('CMS_ACCESS_CMSMain');
                $options = $subsites->toDropdownMap('ID', 'Title');
                $newState->setSubsiteId(join(',', array_keys($options ?? [])));
            }

            $result = $callback($args);
            $newState->setSubsiteId(null);

            return $result;
        });
    }
```

tests/php/BaseSubsiteTest.php
```diff
@@ -15,7 +15,9 @@ protected function setUp(): void
 
         SubsiteState::singleton()->setUseSessions(true);
         Config::modify()->set(Subsite::class, 'write_hostmap', false);
-        Subsite::$force_subsite = null;
+        SubsiteState::singleton()->withState(function (SubsiteState $newState) {
+            $newState->setSubsiteId(null);
+        });
     }
```
sabina-talipova, line 21: And actually, `withState` shouldn't work here as well.

src/Extensions/SiteTreeSubsites.php
sabina-talipova: I'm not sure that I should set subsite ID and after that do the following checking 
```
$subsite = $newState->getSubsiteId() ?: SubsiteState::singleton()->getSubsiteId();
if (empty($this->owner->SubsiteID) || $subsite == $this->owner->SubsiteID) {
...
```
I think this `$subsite == $this->owner->SubsiteID` should return `true` is I set ID.

src/Extensions/SiteTreeSubsites.php
sabina-talipova: DONE

src/Extensions/SiteTreeSubsites.php
emteknetnz: I *think* this block is now supposed to be

```php
return SubsiteState::singleton()->withState(function (SubsiteState $newState) {
    $newState->setSubsiteId($this->owner->SubsiteID);
    return (bool) $this->owner->validURLSegment();
});
```

The whole point is to get a return value from `$this->owner->validURLSegment()` running in the context of the "current" subsite being forced to `$this->owner->SubsiteID`.  All of the deprecated `force_subsite` logic should be replaced by `withState()`

src/Reports/SubsiteReportWrapper.php
emteknetnz: We cannot create a new `beforeAfterQuery()` method, because SubsiteReportWrapper extends the abstract class ReportWrapper which has

```
    /**
     * Override this method to perform some actions prior to querying.
     */
    public function beforeQuery($params)
    {
    }

    /**
     * Override this method to perform some actions after querying.
     */
    public function afterQuery()
    {
    }
```

However looking closer at SubsiteReportWrapper and ReportWrapper, I don't think there's anyway we can replicate the old functionality.

There's also no reference in the codebase to SubsiteReportWrapper, so I'm inclined to just deprecate SubsiteReportWrapper in CMS4 and delete in CMS 5

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-subsites/pull/503

PR #503, opened 2023-01-25
API Remove deprecated code

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10546

SubsiteReportWrapper can be tested with

```php
<?php

use SilverStripe\CMS\Reports\BrokenLinksReport;
use SilverStripe\Subsites\Reports\SubsiteReportWrapper;

class MyReport extends SubsiteReportWrapper
{
    public function title()
    {
        return 'My report';
    }

    public function __construct()
    {
        parent::__construct(BrokenLinksReport::class);
    }
}

```

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\Subsites\Extensions;
 
-use SilverStripe\Dev\Deprecation;
 use Page;
 use SilverStripe\CMS\Forms\SiteTreeURLSegmentField;
 use SilverStripe\CMS\Model\SiteTree;
@@ -10,7 +9,6 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTP;
 use SilverStripe\Core\Config\Config;
-use SilverStripe\Core\Convert;
 use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
@@ -76,13 +74,7 @@ class SiteTreeSubsites extends DataExtension
             return;
         }
 
-        $subsiteID = null;
-        if (Subsite::$force_subsite) {
-            $subsiteID = Subsite::$force_subsite;
-        } else {
-            $subsiteID = SubsiteState::singleton()->getSubsiteId();
-        }
-
+        $subsiteID = SubsiteState::singleton()->getSubsiteId();
         if ($subsiteID === null) {
             return;
         }
@@ -423,7 +415,7 @@ class SiteTreeSubsites extends DataExtension
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
         if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url ?? '');
+            $url = preg_replace('#//[^/]+#', '//' . $this->owner->Subsite()->domain(), $url);
         }
         return $url;
     }
@@ -442,22 +434,6 @@ class SiteTreeSubsites extends DataExtension
         $url = Director::absoluteURL($this->owner->Link($action));
         $link = HTTP::setGetVar('SubsiteID', $this->owner->SubsiteID, $url);
         return $link;
-    }
-
-    /**
-     * This function is marked as deprecated for removal in 5.0.0 in silverstripe/cms
-     * so now simply passes execution to where the functionality exists for backwards compatiblity.
-     * CMS 4.0.0 SiteTree already throws a SilverStripe deprecation error before calling this function.
-     * @deprecated 2.2.0 Use updatePreviewLink() instead
-     *
-     * @param string|null $action
-     * @return string
-     */
-    public function alternatePreviewLink($action = null)
-    {
-        Deprecation::notice('2.2.0', 'Use updatePreviewLink() instead');
-        $link = '';
-        return $this->updatePreviewLink($link, $action);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Reports/SubsiteReportWrapper.php
```diff
@@ -2,32 +2,31 @@
...
     public function parameterFields()
     {
         $subsites = Subsite::accessible_sites('CMS_ACCESS_CMSMain', true);
-        $options = $subsites->toDropdownMap('ID', 'Title');
+        $options = $subsites->map()->toArray();
 
-        $subsiteField = TreeMultiselectField::create(
-            'Subsites',
-            _t(__CLASS__ . '.ReportDropdown', 'Sites'),
+        $subsiteField = DropdownField::create(
```
emteknetnz, line 27: Using DropdownField instead of TreeMultiselectField as we no longer support passing multiple IDs - see https://github.com/silverstripe/silverstripe-subsites/pull/504

src/Extensions/SiteTreeSubsites.php
```diff
@@ -423,7 +415,7 @@ public function alternateAbsoluteLink($action = null)
         // This helps deal with Link() returning an absolute URL.
         $url = Director::absoluteURL($this->owner->Link($action));
         if ($this->owner->SubsiteID) {
-            $url = preg_replace('/\/\/[^\/]+\//', '//' . $this->owner->Subsite()->domain() . '/', $url ?? '');
+            $url = preg_replace('#^(https?://)[^/]+#', '$1' . $this->owner->Subsite()->domain(), $url);
```
ChloeSartorelli: What is the intention of this change? Why are we explicitly requiring a schema vs `//example.com/`?

src/Model/Subsite.php
```diff
@@ -218,7 +192,6 @@ public static function currentSubsiteID()
     public static function changeSubsite($subsite)
     {
         // Session subsite change only meaningful if the session is active.
-        // Otherwise we risk setting it to wrong value, e.g. if we rely on currentSubsiteID.
```
ChloeSartorelli, line 221: I think leaving in the `// Otherwise we risk setting it to wrong value` comment is valuable. It provides extra context as to why we care if the session is active at all.

src/Reports/SubsiteReportWrapper.php
```diff
@@ -53,33 +52,40 @@ public function columns()
...
-        // The user has select a few specific sites
-        if (!empty($params['Subsites'])) {
-            Subsite::$force_subsite = $params['Subsites'];
-
-            // Default: restrict to all accessible sites
+        $subsiteID = 0;
+        if (!empty($params['Subsite'])) {
+            $subsiteID = (int) $params['Subsite'];
+        }
+        if ($subsiteID) {
```
ChloeSartorelli: A subsite ID of `0` means "main site" - we should probably allow selecting the main site explicitly with the dropdown.

tests/php/SubsiteTest.php
```diff
@@ -302,7 +302,7 @@ public function testDefaultDomain()
 
         $this->assertEquals($_SERVER['HTTP_HOST'], singleton(Subsite::class)->PrimaryDomain);
         $this->assertEquals(
-            'http://' . $_SERVER['HTTP_HOST'] . Director::baseURL(),
+            'http://' . $_SERVER['HTTP_HOST'],
```
ChloeSartorelli, line 307: Are we saying the correct behaviour for an absolute base URL is to _exclude_ the actual "baseURL" as according to the Director?
This seems like incorrect behaviour to me.

src/Reports/SubsiteReportWrapper.php
```diff
@@ -53,33 +52,40 @@ public function columns()
         return $columns;
```
ChloeSartorelli, line 54: While we're working on getting this report working we should probably change `Subsite::class` to some translatable string - `SilverStripe\Subsites\Model\Subsite` is not a good column name

src/Reports/SubsiteReportWrapper.php
```diff
@@ -2,32 +2,31 @@
...
 {
-
     /**
      * @return FieldList
      */
     public function parameterFields()
     {
         $subsites = Subsite::accessible_sites('CMS_ACCESS_CMSMain', true);
-        $options = $subsites->toDropdownMap('ID', 'Title');
+        $options = $subsites->map()->toArray();
```
ChloeSartorelli: I think an "all" option should be added, which should be the default value. Given that we've removed the ability to select all subsites (because we can't set multiple subsite IDs) I think this is a reasonable compromise.

To add that ability, you can do this:
```php
public function beforeQuery($params)
{
    if (/*all subsites*/) {
        Subsite::disable_subsite_filter(true);
    }
}
```
and then obviously the reverse in `afterQuery()`.

The option could have, for example, a value of `-1` in the dropdown to make it clearly distinct.

GuySartorelli (changes requested):

src/Extensions/SiteTreeSubsites.php
emteknetnz: Intention was I just found it original confusing, it's a rewrite of a rewrite https://github.com/silverstripe/silverstripe-subsites/pull/499/files#diff-04ef34fc7aa4d36fefe818e18aa1c278aa3f12ff96e4cc9f694f57ea5db2c1f7R429

I've removed the (https?:) bit.  This change is covered by unit tests in SiteTreeSubsitesTest.php

src/Model/Subsite.php
emteknetnz: Added back in

src/Reports/SubsiteReportWrapper.php
emteknetnz: It's currently selectable via the dropdown

tests/php/SubsiteTest.php
emteknetnz: In this context Director::baseURL() is simply '/' - so all we're doing here is updating the unit test to not expect a trailing slash

src/Reports/SubsiteReportWrapper.php
emteknetnz: Updated

src/Reports/SubsiteReportWrapper.php
emteknetnz: Updated

tests/php/SubsiteTest.php
ChloeSartorelli: Right, the _value_ of the test is correct, but the _intent_ of the test appears to have changed - the test _should_ include the base URL if there is one (there just isn't one in this case) - IMO either a new additional test should be added where a base URL is set and then tested against, or simple a comment can be added which says something along the lines of "If there is a Director::baseURL() value this will also be included" just to make it clear what the functionality should be in that case.

emteknetnz: Added comment

GuySartorelli (approved):

PR #504, opened 2023-01-25
API Deprecate passing multiple IDs

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10546

Passing multiple ID's makes for flaky code, it also probably doesn't actually work right now e.g. https://github.com/silverstripe/silverstripe-subsites/blob/2.7/src/Extensions/FileSubsites.php#L95

### DIFF
--- a/src/State/SubsiteState.php
+++ b/src/State/SubsiteState.php
@@ -5,6 +5,7 @@ namespace SilverStripe\Subsites\State;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Resettable;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * SubsiteState provides static access to the current state for subsite related data during a request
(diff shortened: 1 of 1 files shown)


src/State/SubsiteState.php
```diff
@@ -48,6 +49,9 @@ public function getSubsiteId()
      */
     public function setSubsiteId($id)
     {
+        if (!is_int($id) && !is_null($id)) {
+            Deprecation::notice('2.8.0', 'Passing multiple IDs is deprecated, only pass a single ID instead.');
+        }
```
ChloeSartorelli: A string `"1"` would still be valid and not deprecated - or if it would be deprecated, it shouldn't have this message as it's still only a single ID.

GuySartorelli (changes requested):

src/State/SubsiteState.php
emteknetnz: Updated

GuySartorelli (approved):

PR #511, opened 2023-03-01
API Deprecation warning was added to afterQuery method

sabina-talipova:
### Description
`SubsiteReportWrapper::afterQuery()` is marked as deprecated since 2.7 version

### Parent issue
- https://github.com/silverstripeltd/product-issues/issues/699


GuySartorelli (commented):
This method still exists on `ReportWrapper` which this class still subclasses in CMS 5. No deprecation message is needed.
If you agree, we can close the PR. If you think it should be deprecated still please state your reasoning

Issue #509, opened 2023-02-16
Subsites + Fluent: Config setting not working


MLindenhofer:
Hi,

According to the documentation (https://github.com/silverstripe/silverstripe-subsites#using-subsites-in-combination-with-fluent) it's possible to ignore the subsites locale and force using the FluentState locale instead but it seems it does not work anymore. We have the setting in our yml-file but it's ignored and the subsite still uses the locale from the subsite-settings.

Maybe you should update the documentation if it's really ended working :). 

 For our project I've used this workaround in PageController's init()-function:

`            if ($this->dataRecord->SubsiteID !== 0){
                i18n::set_locale(FluentState::singleton()->getLocale());
            }
`

Best regards

maxime-rainville:
Can you clarify what version of Fluent you are using? 

Honestly, this is not a set up we run very often in our own project. I have no idea if what's stated in the doc is even true. If you do find a way to make it work, please report your finding here and/or open a PR to update the doc.

MLindenhofer:
I think the fluent version is not cause of the problem (in our project it was Fluent 5.1.1 ), the code-part where the old yml-config-var seems not to be available anymore. Searched for it. Nothing found. 

Did my very first PR with the doc update, to help other developers with the same problem.

PR #514, opened 2023-03-07
Update documentation "Using Subsites in combination with Fluent" part

MLindenhofer:
References: #509 - the yml-config solution for setting the Fluent locale instead of the Subsite-locale is not existing anymore. I replaced the old solution with a workaround which works for me. 

it's my first PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ or by defining the subsiteCMSShowInMenu function in your admin:
 
 When using Subsites in combination with Fluent module, the Subsites module sets the i18n locale to the language defined in the current Subsite. When this behaviour is not desired and you need to use the locale in FluentState you can set the current Fluent locale in your PageController like this:
 
+*PageController.php*
 ```
 :::php
      protected function init(){
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ permissions" will imply that the person will likely be able to escalate his/her 
 
 For user documentation please see:
 
-1. [Setting up subsites](docs/en/userguide/set_up.md)
-1. [Working with subsites](docs/en/userguide/working_with.md)
+ 1. [Setting up subsites](docs/en/userguide/set_up.md)
+ 1. [Working with subsites](docs/en/userguide/working_with.md)
 
 ## Features & limitations
 
@@ -134,7 +134,7 @@ themes:
   - '$default'
 ```
 
-You may also completely define your own cascading theme lists for CMS users to choose as theme options for their
+You may also completely define your own cascading theme lists for CMS users to choose as theme options for their 
 subsite:
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -22,38 +22,38 @@ permissions" will imply that the person will likely be able to escalate his/her 
 
 For user documentation please see:
 
- 1. [Setting up subsites](docs/en/userguide/set_up.md)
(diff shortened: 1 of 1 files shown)


README.md
ChloeSartorelli, line 1: Please revert the unrelated style changes you have made to this file. Those changes aren't related to the purpose of this pull request.

README.md
```diff
@@ -220,11 +220,16 @@ or by defining the subsiteCMSShowInMenu function in your admin:
...
 
-When using Subsites in combination with Fluent module, the Subsites module sets the i18n locale to the language defined in the current Subsite. When this behaviour is not desired and you need to use the locale in FluentState use the following setting in your yml config file:
+When using Subsites in combination with Fluent module, the Subsites module sets the i18n locale to the language defined in the current Subsite. When this behaviour is not desired and you need to use the locale in FluentState you can set the current Fluent locale in your PageController like this:
 
-```yaml
-    SilverStripe\Subsites\Extensions\SiteTreeSubsites:
-      ignore_subsite_locale: true
+*PageController.php*
+```
+:::php
```
ChloeSartorelli:
```suggestion
```php
```

README.md
```diff
@@ -220,11 +220,16 @@ or by defining the subsiteCMSShowInMenu function in your admin:
...
+     protected function init(){
```
ChloeSartorelli:
```suggestion
     protected function init()
     {
```

README.md
```diff
@@ -220,11 +220,16 @@ or by defining the subsiteCMSShowInMenu function in your admin:
...
+        if ($this->dataRecord->SubsiteID !== 0){
+            i18n::set_locale(FluentState::singleton()->getLocale());
+        }
+     }
 ```
```
ChloeSartorelli: Can you please explain why you're making this change? What was wrong with the yaml configuration documented here?

GuySartorelli (changes requested):
Hi,

Thank you for your contribution, and sorry it's taken so long to get back to you. There are some requested changes and clarification questions below.

README.md
MLindenhofer: I make this change, because the yaml configuration documented is not in use in the Subsite-Module. So even with the yml configuration set, the i18n-Locale is still the one defined for the subsite and not the one from fluent. My solution is just one possible workaround.

README.md
MLindenhofer: Done :)

README.md
ChloeSartorelli: Please make this change, so that the documentation is showing best practice.

README.md
ChloeSartorelli: > I make this change, because the yaml configuration documented is not in use in the Subsite-Module. 

That yaml configuration is used here:
https://github.com/silverstripe/silverstripe-subsites/blob/4e292b8e89f9ee214340a215285cdcb526a2b5f8/src/Extensions/SiteTreeSubsites.php#L399-L417

MLindenhofer: ah I see - okay. In module version 2.6.0 it wasn't there :). Seems it changed in the meanwhile. 
Than my change is unnecessary of course .

ChloeSartorelli: Okay cool. I'll close this PR then. Thank you for your efforts and patience though, it's definitely appreciated.

PR #516, opened 2023-03-17
Add subsite change detection

### LINKED ISSUE #515: new feature: subsite change detector
I've recently played around with a little js snippet and I think it would be a worthwhile addition to the module

the goal is to detect when a subsite change has occurred in another tab. This can help preventing many odd situations where you would click and suddenly see a whole new page due to the context change

the solution works like this, it's a really simple js snippet
- store the currently selected subsite in local storage
- detect changes in storage. if subsite has changed, prompt for reload

what do you think?

```js
  // Subsite change detector
  var subsiteSessionKey = "admin_subsite_id";
  var subsiteSelectedId = null;
  window.addEventListener("storage", () => {
    var tabId = localStorage.getItem(subsiteSessionKey);
    if (tabId && subsiteSelectedId != tabId) {
      if (confirm("The current subsite has changed. Reload the page ?")) {
        window.location.reload();
      }
    }
  });
  $("#SubsitesSelect").entwine({
    onmatch: function () {
      subsiteSelectedId = this.find("option[selected]").attr("value");
      localStorage.setItem(subsiteSessionKey, subsiteSelectedId);
    },
  });
```

## Acceptance criteria
- Users are warned when their session is changed to target a different subsite.
- Validate what currently happens when the user tries to make a change when looking at subsite that doesn't match what is in their session.
  - Clarify what should happen when the user chooses to ignore your warnings.
- Need to make sure the Subsite ID in the PHP session is always respected even if it doesn't match the local storage Susbsite ID.

## PR
- https://github.com/silverstripe/silverstripe-subsites/pull/516

lekoala:
Closes https://github.com/silverstripe/silverstripe-subsites/issues/515


GuySartorelli:
This looks really useful, thanks!
As a new feature, this will need to target the `3` branch. Please retarget and rebase this PR.

GuySartorelli:
@lekoala I've retargetted this at the `3` branch and made the necessary changes for it to pass the new linting rules.

However, while testing this I found a (probably very common) scenario where this will cause _more_ confusion rather than less.
When editing a page in tab A in subsite A, after changing subsites to subsite B in tab B, tab A tells me it wants me to refresh because I changed subsites. But refreshing tab A sets the subsite back to subsite A - so it appears I haven't made any change at all - and then _tab B_ tells me I should refresh the page because I changed subsites again!

Can you think of an easy way to resolve this in a general way? We could obviously check if someone's in a cms page edit form, but that seems like a narrow scope when this scenario could theoretically happen in other contexts as well.

lekoala:
@GuySartorelli reloading should load the proper website no? any way, it would be easy i think to pass along the SubsiteID in the url to force loading the last active subsite

michalkleiner:
The problem we had in past was that not all model admins or all CMS UI was the same for all subsites, so where should it take you on a reload when you were editing a page type that is not available for the destination subsite? Go to default pages view?

lekoala:
@michalkleiner fair point. i guess that would lead to a 404 so i'm not even sure how the redirect would work. i'm not saying my approach in 100% bulletproof (it works for me, and that's already something :) ) so maybe that's why it should be opt in?

GuySartorelli:
I'd hesitate to merge this even as an opt-in feature at the moment, since it doesn't play nicely with editing pages which is arguably the most common action in the CMS.

michalkleiner:
I would be ok to have this opt in, if it always took you to just /admin and it clearly said it in the dialog. Or even better with the message customisable, which would require adjustments to source it from config and put it e.g. in a data attribute, but that's all doable.

GuySartorelli:
@lekoala Are you still interested in having this PR merged in? If the changes Michal suggested in the last comment are made we can accept it.

lekoala:
still interested but with limited time to spend on it

GuySartorelli:
I'm going to close this for now since there has been no action for several months - we can reopen it when/if you get time to pick it up again.

lekoala:
no worries ;) it's just it's working "good enough" for me at the moment so i have limited interest in getting this further

PR #518, opened 2023-03-27
FIX Don't use deprecated API

GuySartorelli:
`Deprecation::notification_version()` is deprecated - we shouldn't be calling it.
There's a `_config` directory so we don't need to keep an empty `_config.php` file.

## Issue
- https://github.com/silverstripeltd/product-issues/issues/692

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,5 +1,0 @@
-<?php
-
-use SilverStripe\Dev\Deprecation;
-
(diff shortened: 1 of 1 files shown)


GuySartorelli:
behat failures are pre-existing and aren't related to this PR

maxime-rainville (approved):
LGTM

maxime-rainville:
Behat failure unrelated.

PR #527, opened 2023-06-21
Add extension to correctly support element preview

michalkleiner:
As per the title, conditional based on base element class existing.

Tried to get the subsite ID from the element, if it has the model extension adding it applied, or from its page.

Targeting 2.8 as I see this as a fix for a broken preview mechanism rather than a new feature or an enhancement, but ok to target 3.0 as well since this can easily be done in project code until this is released.


GuySartorelli (approved):
Nice and clean, works well locally.
I agree this is a bug fix that can and should be released as a patch.

GuySartorelli:
Released as [2.8.2](https://github.com/silverstripe/silverstripe-subsites/releases/tag/2.8.2) and [3.0.2](https://github.com/silverstripe/silverstripe-subsites/releases/tag/3.0.2) - and congrats, this is apparently your first contribution to subsites 🥳

PR #533, opened 2023-08-16
don't trigger permissionFailure when it's not needed

### LINKED ISSUE #529: Incorrect template shown when user doesn’t have CMS access (specifically when the subsites module is installed)
Related issue https://github.com/silverstripe/silverstripe-admin/issues/1443

Due to permissions failure being thrown with Security::permissionFailure($this->owner), incorrect template is applied and a user with no access can see the admin layout

I think this is probably more a framework level fix, but raising the issue here as well

PR
---

https://github.com/silverstripe/silverstripe-subsites/pull/533

lekoala:
Fixes https://github.com/silverstripe/silverstripe-subsites/issues/529

LeftAndMain already has a canView check that will use under the hood the alternateAccessCheck
Not trigger the permissionFailure allows proper handling of the error instead of showing a ever loading logo in admin v5

### DIFF
--- a/src/Extensions/LeftAndMainSubsites.php
+++ b/src/Extensions/LeftAndMainSubsites.php
@@ -369,7 +369,7 @@ class LeftAndMainSubsites extends LeftAndMainExtension
             }
 
             // We have not found any accessible section or subsite. User should be denied access.
-            return Security::permissionFailure($this->owner);
+            // This is handled already by LeftAndMain thanks to alternateAccessCheck
         }
 
         // Current site is accessible. Allow through.


GuySartorelli:
My understanding is that the underlying issue this aims to fix is a problem for CMS 4 as well - please retarget this PR to the `2.8` branch so that we can resolve the issue for both major releases.

lekoala:
done

GuySartorelli (approved):
Nice, very elegant and works great. Thank you for contributing this!

GuySartorelli:
This change will be automatically tagged once CI has finished running on the branch, and will be automatically merged up to and tagged in `3.0` some time after that (on a cron).

Issue #535, opened 2023-08-23
CMSProfileController required_permission_codes breaks with subsites


emteknetnz:
https://github.com/silverstripe/silverstripe-admin/pull/1027 caused failures when subsites was installed on the recipe-kitchen sink build here https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/5892574918/job/15998642199

There was some [initial investigation](https://github.com/silverstripe/silverstripe-admin/pull/1555#issuecomment-1689037248) done that there's some additional work to be done on the subsites module to get it compatible

## PRs
- https://github.com/silverstripe/silverstripe-subsites/pull/536

GuySartorelli:
Merged PR

PR #536, opened 2023-08-24
FIX Set extraCodes to false for CMSProfileController

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-subsites/issues/535

### DIFF
--- a/src/Extensions/LeftAndMainSubsites.php
+++ b/src/Extensions/LeftAndMainSubsites.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Subsites\Extensions;
 
 use SilverStripe\Admin\AdminRootController;
 use SilverStripe\Admin\CMSMenu;
+use SilverStripe\Admin\CMSProfileController;
 use SilverStripe\Admin\LeftAndMainExtension;
 use SilverStripe\CMS\Controllers\CMSPagesController;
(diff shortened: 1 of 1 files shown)


src/Extensions/LeftAndMainSubsites.php
```diff
@@ -88,7 +89,13 @@ public function sectionSites($includeMainSite = true, $mainSiteTitle = 'Main sit
         // Collect permissions - honour the LeftAndMain::required_permission_codes, current model requires
         // us to check if the user satisfies ALL permissions. Code partly copied from LeftAndMain::canView.
         $codes = [];
-        $extraCodes = Config::inst()->get(get_class($this->owner), 'required_permission_codes');
+        $ownerClass = get_class($this->owner);
```
michalkleiner, line 92: You can also use `$ownerClass` on row 96 (old)/103 (new).

GuySartorelli (commented):
Looks good and in a local kitchen sink install it passes the test that was failing.

I'll give you a chance to make the change that Michal has suggested, though that's not a deal breaker I think.

src/Extensions/LeftAndMainSubsites.php
emteknetnz: Upated

GuySartorelli (approved):
LGTM👍

PR #547, opened 2023-10-12
MNT Remove unused Translatable

sabina-talipova:
## Description
The Translatable module is no longer in use.
### Related issue
- https://github.com/silverstripe/silverstripe-framework/issues/10548


GuySartorelli (changes requested):
This should either target `3.0` if it's to be released as a patch, or `3` if it's to be included in the next minor release. `3.1` is in a change freeze (as it has had the RC tagged) so non-patchable changes shouldn't be targetting that branch.

GuySartorelli (approved):
Looks good to me.

Issue #561, opened 2024-03-16
SiteTreeSubsites potentially inserts CopyToSubsiteID multiple times


sephyhe:
### Module version(s) affected

^3.0

### Description

In the file - SiteTreeSubsites.php, the function `updateCMSFields` inserts SubsiteOperations with `CopyToSubsiteID`, `CopyToSubsiteWithChildren` and form action `copytosubsite`.
If there is a new page extended from Page and contains code `$this->extend('updateCMSFields', $fields);`
```
class ParentPage extends Page
{
    private static $table_name = 'ParentPage';

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        //Some changes
        ...

        $this->extend('updateCMSFields', $fields);
        return $fields;
    }
}
```

And all the other pages extended from ParentPage will get an issue:
```[Emergency] Uncaught RuntimeException: SilverStripe\Forms\{closure}() I noticed that a field called 'CopyToSubsiteID' appears twice```

Because the code here(https://github.com/silverstripe/silverstripe-subsites/blob/0fe63e92bd9a360471880bf4849f18de612aa0c1/src/Extensions/SiteTreeSubsites.php#L125) try to insert SubsiteOperations more than one time.

Therefore, it will be good to check if the fields are exiting or not before inserting them.

### How to reproduce

```
class ParentPage extends Page
{
    private static $table_name = 'ParentPage';

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        //Some changes
        ...

        $this->extend('updateCMSFields', $fields);
        return $fields;
    }
}
```

```
class ChildPage extends ParentPage
{
     private static $table_name = 'ChildPage';

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        // Error will happen at this page
        ...
        return $fields;
    }
}
```

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

GuySartorelli:
This is not a bug with subsites, it is a problem with how `$this->extend()` is being used.
If your subclass and parent class both call `$this->extend('updateCMSFields')` then it is expected that the same method will be called twice, and all of its code will be executed twice.

I recommend that instead you use `beforeUpdateCMSFields()` in your page subclasses.
[...]

PR #562, opened 2024-03-17
Check the Form fields before inserting them to avoid duplication Error

sephyhe:
## Description
Check the Form fields before inserting them to avoid duplication Error

## Issues
- #561


GuySartorelli:
As mentioned in the linked issue, this is not a bug with subsites. I will close this PR as it addresses the problem from the wrong angle.

PR #576, opened 2024-06-10
MNT Fix scss lint warnings

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/228

### DIFF
--- a/client/src/styles/LeftAndMain_Subsites.scss
+++ b/client/src/styles/LeftAndMain_Subsites.scss
@@ -13,19 +13,13 @@
   height: 51px;
   border-bottom: 3px solid #d4d0c8;
   color: #fff;
-}
 
-#SubsiteActions {
   fieldset {
     padding: 3px;
     border-style: none;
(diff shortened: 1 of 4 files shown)


client/src/styles/LeftAndMain_Subsites.scss
```diff
@@ -75,9 +69,7 @@
   .field {
     margin-left: 100px;
   }
-}
 
-#Form_AddSubsiteForm {
```
emteknetnz, line 80: Why is this ID being removed?

emteknetnz (changes requested):

client/src/styles/LeftAndMain_Subsites.scss
ChloeSartorelli: It's not being removed - this block with a duplicate selector is being merged in with the above block which has the exact same selector. The dist result is identical.

emteknetnz (approved):

PR #584, opened 2024-07-30
API Remove GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799

### DIFF
--- a/src/Middleware/InitStateMiddleware.php
+++ b/src/Middleware/InitStateMiddleware.php
@@ -23,7 +23,6 @@ class InitStateMiddleware implements HTTPMiddleware
      */
     private static $admin_url_paths = [
         'dev/',
-        'graphql/',
     ];
 
     public function process(HTTPRequest $request, callable $delegate)
--- a/tests/behat/features/operate-subsites.feature
+++ b/tests/behat/features/operate-subsites.feature
@@ -68,6 +68,7 @@ Feature: Create and select a subsite
     When I go to "admin/assets"
     And I select "My subsite" from "SubsitesSelect"
     And I press the "Add folder" button
+    And I wait for 2 seconds
     And I fill in "Folder name" with "sfol"
     And I press the "Create" button
     When I go to "admin/assets"
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "dependencies": {},
   "devDependencies": {
     "@silverstripe/eslint-config": "^1.3.0",
-    "@silverstripe/webpack-config": "^2.1.0",
+    "@silverstripe/webpack-config": "^3.0.0-alpha1",
     "webpack": "^5.74.0",
     "webpack-cli": "^5.0.0"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -1357,10 +1357,10 @@
     stylelint-config-standard "^36.0.0"
     stylelint-scss "^6.2.1"
 
-"@silverstripe/webpack-config@^2.1.0":
-  version "2.1.0"
-  resolved "https://registry.yarnpkg.com/@silverstripe/webpack-config/-/webpack-config-2.1.0.tgz#9d6409498126db766299fa40f209f618515932a4"
-  integrity sha512-8rE2K3AXkIP6crDTBXdHz/3hiNVp2XpINkWGDBKzvMS6sGRcVTluPs12BdOXYo1aUZEeqDk9QYOLLGGgZxnOdw==
+"@silverstripe/webpack-config@^3.0.0-alpha1":
+  version "3.0.0-alpha1"
+  resolved "https://registry.yarnpkg.com/@silverstripe/webpack-config/-/webpack-config-3.0.0-alpha1.tgz#9576b8b1deac92ea96636e149b051b513ce14ab6"
+  integrity sha512-xBsiewBVAUIGdshWSs8ChAg4JXNNpHMwBLhjhYJCPFKe2HpNx3OqFEIqjR9YgSbCVSxkNh4x5yvYklu6+k+YpA==
   dependencies:
     "@babel/core" "^7.19.6"
     "@babel/preset-env" "^7.19.4"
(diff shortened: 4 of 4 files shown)


tests/behat/features/operate-subsites.feature
```diff
@@ -28,7 +28,7 @@ Feature: Create and select a subsite
     And I press the "Add folder" button
     And I select "Main site" from "SubsitesSelect"
     # Using a short folder name so that it doesn't get truncated on the frontend
-    And I fill in "Folder name" with "mfol"
+    And I fill in "Form_folderCreateForm_Name" with "mfol"
```
emteknetnz: For some reason we now need to get the same field via ID rather than name

src/Middleware/InitStateMiddleware.php
```diff
@@ -23,7 +23,6 @@ class InitStateMiddleware implements HTTPMiddleware
      */
     private static $admin_url_paths = [
         'dev/',
-        'graphql/',
```
NightJar, line 26: You need a hook in whatever consumes this config in order to prevent issues that would block people who _do_ use GraphQL with e.g. custom features in their CMS.

emteknetnz: It's private static config, people can update it via their project yml

tests/behat/features/operate-subsites.feature
```diff
@@ -25,10 +25,13 @@ Feature: Create and select a subsite
 
     # Add a file to the main site
     When I go to "admin/assets"
-    And I press the "Add folder" button
     And I select "Main site" from "SubsitesSelect"
+    # Ensure that the new subsite is located before adding folders
+    And I go to "admin/assets"
+    And I press the "Add folder" button
+    And I wait for 2 seconds
```
ChloeSartorelli: Do you remember why this change was necessary? Feels like it could be a regression or some underlying bug we're just testing around?

tests/behat/features/operate-subsites.feature
ChloeSartorelli: Uhhh that feels bad. Did the name of the form field change or something? This feels like it could lead to unexpected regressions, e.g. if someone has js or css relying on form field names?

tests/behat/features/operate-subsites.feature
```diff
@@ -65,10 +68,11 @@ Feature: Create and select a subsite
     Then I should see "file1"
 
     # Add a file to the subsite
-    When I go to "admin/assets"
-    And I select "My subsite" from "SubsitesSelect"
+    When I select "My subsite" from "SubsitesSelect"
+    And I go to "admin/assets"
```
ChloeSartorelli: Is this change for the same reason as the one with "Ensure that the new subsite is located before adding folders" above?

GuySartorelli (changes requested):

tests/behat/features/operate-subsites.feature
emteknetnz: I cannot remember

tests/behat/features/operate-subsites.feature
emteknetnz: I think so

tests/behat/features/operate-subsites.feature
emteknetnz: I don't think it changed, behat was just fussy for some reason

ChloeSartorelli: If nothing changed, nothing needs to change here.

But evidently you've had to change something here, which means something _did_ change. Behat being fussy doesn't explain it, because this change isn't necessary in the current `4` branch, it's only needed when applying the changes in these PRs.

So _something_ you've changed has caused this to need to change, and that's scaring me a little that we don't know what it was or what other side effects that might have.

GuySartorelli (changes requested):

tests/behat/features/operate-subsites.feature
emteknetnz: I'll try changing it back and see if it's resolved, does work locally, we'll see if CI likes it

emteknetnz: Works with it reverted on latest kitchen sink run

tests/behat/features/operate-subsites.feature
ChloeSartorelli: Tried reverting this locally and it passed. Please revert this as it doesn't seem to be necessary.

tests/behat/features/operate-subsites.feature
ChloeSartorelli: Tried reverting this locally and it passed. Please revert this as it doesn't seem to be necessary (keep the "wait for 2 seconds" below, which does seem to be necessary)

GuySartorelli (changes requested):

emteknetnz:
Have reverted

tests/behat/features/operate-subsites.feature
```diff
@@ -1,4 +1,4 @@
-@javascript
+@javascript @sboyd
```
ChloeSartorelli:
```suggestion
@javascript
```

tests/behat/features/operate-subsites.feature
```diff
@@ -25,6 +25,7 @@ Feature: Create and select a subsite
 
     # Add a file to the main site
     When I go to "admin/assets"
+    And I select "Main site" from "SubsitesSelect"
```
ChloeSartorelli:
```suggestion
```
That's being done after "add folder" two lines down.

GuySartorelli (changes requested):

tests/behat/features/operate-subsites.feature
emteknetnz: Removed

tests/behat/features/operate-subsites.feature
emteknetnz: Removed

GuySartorelli (approved):
LGTM

PR #590, opened 2024-08-13
FIX Correct trailing slash for subsite in another domain

GuySartorelli:
Fixes https://github.com/silverstripe/silverstripe-subsites/actions/runs/10278313787/job/28441800184 which consists of failures like the below:

> ```text
> Failed asserting that two strings are identical.
> --- Expected
> +++ Actual
> @@ @@
> -'http://two.mysite.com/'
> +'http://two.mysite.com/'
> ```

As of silverstripe/framework 5.2.20, `Controller::normaliseTrailingSlash()` won't normalise trailing slash for absolute URLs where the domain doesn't match the current site domain.

This means if getting a link to a page on Subsite A from Subsite B, if the domains don't match, the trailing slash won't be normalised.

It's possible that it would be better to implement the `updateIsSiteUrl()` extension hook in an extension for `Director` and saying "yes that's a site URL" if it belongs to a domain of any subsite. I didn't attempt that because I don't know what sort of flow-on effects that might have. I'm especially hesitant to do that in a patch release.

Note that the same problem happened on Fluent, see https://github.com/tractorcow-farm/silverstripe-fluent/pull/872
Most likely we'll want to implement the `updateIsSiteUrl()` extension hook for both modules in a minor release, but I want at least one other opinion before I do that. Either way, IMO this PR should be merged for a patch fix.

## Issue
- https://github.com/silverstripe/.github/issues/290

### DIFF
--- a/src/Model/SubsiteDomain.php
+++ b/src/Model/SubsiteDomain.php
@@ -217,9 +217,11 @@ class SubsiteDomain extends DataObject
      */
     public function absoluteBaseURL()
     {
-        return Controller::join_links(
+        $slash = Controller::config()->get('add_trailing_slash') ? '/' : '';
+        $baseURL = Controller::join_links(
             $this->getAbsoluteLink(),
             Director::baseURL()
         );
+        return (rtrim($baseURL, '/')) . $slash;
     }
 }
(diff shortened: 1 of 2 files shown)


GuySartorelli:
@silverstripe/core-team if anyone has a sec to look at this and give an opinion about the `updateIsSiteUrl()` option for a more robust solution in a minor release, I'd be grateful.

emteknetnz (approved):

PR #593, opened 2024-08-20
API Replace Extension subclasses

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11050

### DIFF
--- a/docs/en/technical.md
+++ b/docs/en/technical.md
@@ -9,7 +9,7 @@ page to a particular subsite.
 
 The subsite module adds functionality to the admin section of the site to allow you to create new subsites and copy 
 pages between the main site and any subsites.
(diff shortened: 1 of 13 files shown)


GuySartorelli (commented):
LGTM, will hold this one off for last since it has failing tests. Will rerun CI after everything else is merged.

GuySartorelli:
CI still failing - I would have said these look unrelated, except the full kitchen sink run was passing so I'd expect this to pass now too.

emteknetnz:
Have fixed up the unit tests

Previously there was a fairly point array assertion on extraStatics(), which was a pointless method on DataExtension. I'm not sure why it was passing in sink CI

There was also a failure based on the recently merged protected extensions PR, which I've updated to use the equivalent public API instead

GuySartorelli (approved):
LGTM

PR #597, opened 2024-09-06
API Update API to reflect changes to CLI interaction

GuySartorelli:
Reflects changes in https://github.com/silverstripe/silverstripe-framework/pull/11353

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11341

PR #599, opened 2024-09-10
API Deprecate API that will be removed

GuySartorelli:
See https://github.com/silverstripe/silverstripe-subsites/pull/597

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11341


emteknetnz (changes requested):
Should change the deprecation notice in all log() methods in all PRs to something along the lines of "Will be replaced with something that symfony does in future verions"

GuySartorelli:
Updated

emteknetnz (approved):

PR #611, opened 2024-11-12
API Make SubsiteXHRController a subclass of AdminController

GuySartorelli:
When I did https://github.com/silverstripe/silverstripe-admin/issues/1761 I thought `SubsiteXHRController` needed form schema logic, but it turns out it doesn't, so it can be a direct subclass of `AdminController`.

There's two PRs here:
1. Fixes a bug where the subsite selector didn't update when creating or updating subsites (found when trying to test this PR). Not doing as a patch because it moves logic into different selectors and who knows what weird nonsense someone might be doing that relies on that. Easier to just fix in 6 since that's where I'm already working.
2. Make `SubsiteXHRController` a subclass of `AdminController`

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762


client/src/js/LeftAndMain_Subsites.js
```diff
@@ -11,16 +11,6 @@
       }
     });
 
-    /*
-     * Reload subsites dropdown when links are processed
-     */
-    $('.cms-container .cms-menu-list li a').entwine({
```
ChloeSartorelli, line 17: Not only is this selector wrong (should be `.cms-menu__list`) but this just straight up isn't necessary. Clicking on the links in the left menu reloads the whole left menu anyway.

client/src/js/LeftAndMain_Subsites.js
```diff
@@ -32,11 +22,15 @@
...
-        $('.cms-container').loadFragment('admin/subsite_xhr', 'SubsiteList');
+    $('#Form_ItemEditForm').entwine({
+      onaftersubmitform(e) {
+        // Only load the fragment if this form is the subsite form.
+        // We can't add a selector to the form itself so check for the tab we added a specific class to.
+        if (this.find('[role="tab"].subsite-model').length > 0) {
+          $('.cms-container').loadFragment('admin/subsite_xhr', 'SubsiteList');
+        }
         this._super(e);
       }
```
ChloeSartorelli, line 35: I think this was trying to reload the selector any time it saw the root.main tab in the subsite edit form.
The selector was wrong, but the approach was wrong too - we should only reload the selector after the form is submitted, not just any time the form loads.

src/Controller/SubsiteXHRController.php
```diff
@@ -2,20 +2,25 @@
...
-class SubsiteXHRController extends LeftAndMain
+class SubsiteXHRController extends AdminController
 {
     private static $url_segment = 'subsite_xhr';
 
-    private static $ignore_menuitem = true;
+    public function index(HTTPRequest $request): HTTPResponse
+    {
+        return $this->getResponseNegotiator()->respond($request);
+    }
```
ChloeSartorelli, line 25: This is necessary because we're using `loadFragment` in js rather than directly hitting the `SubsiteList` method as a controller action

src/Controller/SubsiteXHRController.php
```diff
@@ -36,33 +41,21 @@ public function canView($member = null)
...
-        $negotiator->setCallback('SubsiteList', function () {
-            return $this->SubsiteList();
-        });
-
-        return $negotiator;
+        return new PjaxResponseNegotiator([
+            'SubsiteList' => function () {
+                return $this->SubsiteList();
+            },
+        ]);
```
ChloeSartorelli, line 54: That's the only fragment this controller ever cares about.
Using `new` because that's what's done everywhere that uses it right now and it's not an `Injectable` class.

src/Extensions/LeftAndMainSubsites.php
```diff
@@ -45,6 +47,42 @@ protected function onInit()
...
+            'SubsiteSwitchList',
+        ];
+    }
+
+    /**
+     * Generates a list of subsites with the data needed to
+     * produce a dropdown site switcher
+     * @return SS_List<Subsite>
+     */
+    public static function SubsiteSwitchList(): SS_List
```
ChloeSartorelli: Since `SubsiteXHRController` isn't a subclass of `LeftAndMain` anymore, it doesn't get access to the methods in this extension. This method is only needed in templates anyway so I've just made it _always_ available in templates.

Since I had to break the method signature to make it static I took the liberty of renaming it to make it much clearer what its purpose is.

The alternative would be to make a separate extension with just this logic and apply it to both `LeftAndMain` and the xhr controller. Happy to do that if you prefer that approach.

templates/SilverStripe/Admin/LeftAndMain_Menu.ss
```diff
@@ -5,7 +5,8 @@
         <% include SilverStripe\\Admin\\LeftAndMain_MenuLogo %>
         <% include SilverStripe\\Admin\\LeftAndMain_MenuStatus %>
 
-        <% if $ListSubsites.Count > 1 %>
+        <% if $SubsiteSwitchList.Count > 1 %>
+            <% require javascript('silverstripe/subsites:client/dist/js/LeftAndMain_Subsites.js') %>
```
ChloeSartorelli, line 9: Moved the require code here instead of in the method since we don't need to re-require it for the xhr requests and because frankly a require call in an arbitrary PHP method is just gross

src/Controller/SubsiteXHRController.php
```diff
@@ -36,33 +43,21 @@ public function canView($member = null)
     }
 
     /**
-     * Allow access if user allowed into the CMS at all.
+     * Get a Pjax response negotiator for the subsite list
      */
-    public function canAccess()
```
ChloeSartorelli, line 43: `canAccess()` is a non-standard method that's only applied to `LeftAndMain` classes via the `LeftAndMainSubsites` extension. It no longer applies here, so we use `required_permission_codes` instead set to the less restrictive of these since `canView()` is handling the main access check for us.

src/Service/ThemeResolver.php
```diff
@@ -41,7 +41,7 @@ class ThemeResolver
      * @param Subsite $site
      * @return array
      */
-    public function getThemeList(Subsite $site)
+    public function getThemeList(Subsite $site): array
```
ChloeSartorelli: This resolves 
```text
1) SilverStripe\Subsites\Tests\SiteTreeSubsitesTest::testThemeResolverIsUsedForSettingThemeList
TypeError: SilverStripe\View\SSViewer::set_themes(): Argument #1 ($themes) must be of type array, null given, called in /var/www/html/vendor/silverstripe/subsites/src/Extensions/SiteTreeSubsites.php on line 405
```

The param in `SSViewer::set_themes()` is typed to require an array specifically, so this should return an array specifically.

emteknetnz (approved):

GuySartorelli:
rebased

emteknetnz (approved):

PR #612, opened 2024-11-14
API Ensure themelist returns an array

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/328

Fixes https://github.com/silverstripe/silverstripe-subsites/actions/runs/11697085928/job/32899670628

`1) SilverStripe\Subsites\Tests\SiteTreeSubsitesTest::testThemeResolverIsUsedForSettingThemeList
TypeError: SilverStripe\View\SSViewer::set_themes(): Argument #1 ($themes) must be of type array, null given, called in /home/runner/work/silverstripe-subsites/silverstripe-subsites/src/Extensions/SiteTreeSubsites.php on line 405`

### DIFF
--- a/src/Extensions/LeftAndMainSubsites.php
+++ b/src/Extensions/LeftAndMainSubsites.php
@@ -17,7 +17,7 @@ use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\HiddenField;
 use SilverStripe\Model\List\ArrayList;
 use SilverStripe\ORM\DataObject;
-use SilverStripe\ORM\SS_List;
+use SilverStripe\Model\List\SS_List;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
I prefer https://github.com/silverstripe/silverstripe-subsites/pull/611/files#r1838742589 as it feels cleaner and ensures we do always set the theme list, whereas it looks like this change could result in having an old not-relevant theme list if `$themelist` is null.

emteknetnz:
That linked solution should still fail as it'll attempt to return null in some circumstances. This unit test started failing after the template work was merged in I think - possibly the linked solution will also fail after rebasing top of the latest work in 6? Could you double check to see if that's the case?

GuySartorelli:
It's fixing literally this exact same error in that PR, so no it won't still fail.

emteknetnz:
Test locally, it does fix it. Will update

GuySartorelli (approved):
LGTM

PR #621, opened 2024-12-16
MNT Fix unit test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/349

Fixes https://github.com/silverstripe/silverstripe-subsites/actions/runs/12342588552/job/34442723479#step:12:109

```
1) SilverStripe\Subsites\Tests\GroupSubsitesTest::testAlternateTreeTitle
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'The A Team <i>(global group)</i>'
+'The A Team &lt;i&gt;(global group)&lt;/i&gt;'
```

### DIFF
--- a/src/Extensions/GroupSubsites.php
+++ b/src/Extensions/GroupSubsites.php
@@ -126,6 +126,17 @@ class GroupSubsites extends Extension implements PermissionProvider
         }
     }
 
+    public function getTreeTitle(): string
+    {
+        // Overrides Hierachy::getTreeTitle() to change where Convert::raw2xml() is called
+        // so that the html added updateTreeTitle() is not escaped.
+        $owner = $this->getOwner();
+        $title = $owner->MenuTitle ?: $owner->Title;
+        $title = Convert::raw2xml($title);
+        $owner->extend('updateTreeTitle', $title);
+        return $title;
+    }
+
     /**
      * If this group belongs to a subsite, append the subsites title to the group title to make it easy to
      * distinguish in the tree-view of the security admin interface.


GuySartorelli (changes requested):
Are we sure this behaviour is correct, and isn't indicative of an unintended regression? When changing a test to match a change in behaviour I'd expect to see some explanation of why the behavioural change is correct.

emteknetnz:
The unit test is simply calling Hierachry::getTreeTitle() which has recently been updated

```php
    public function getTreeTitle(): string
    {
        $owner = $this->getOwner();
        $title = $owner->MenuTitle ?: $owner->Title;
        $owner->extend('updateTreeTitle', $title);
        return Convert::raw2xml($title ?? '');
    }
```

GuySartorelli:
Right, I get that, but _should_ `raw2xml` be applied there? It looks like it wasn't in the past or this test wouldn't have failed. Does this cause peoples' legitimate HTML to not render correctly in some context? If so that's an unintended regression.

I believe at the time I tested groups in `TreeDropdownField` _without subsites_ but it appears that `GroupSubsites::updateTreeTitle()` adds some _intentional HTML_ which should either not be added if it doesn't display as expected in context that calls `getTreeTitle()` - or if it would display correctly, it should be permitted instead of being escaped (i.e. `raw2xml` called on the title, not on the updated result)

emteknetnz:
Updated

src/Extensions/GroupSubsites.php
```diff
@@ -126,6 +126,17 @@ protected function updateCMSFields(FieldList $fields)
...
+    public function getTreeTitle(): string
+    {
+        // Overrides Hierachy::getTreeTitle() to change where Convert::raw2xml() is called
+        // so that the html added updateTreeTitle() is not escaped.
+        $owner = $this->getOwner();
+        $title = $owner->MenuTitle ?: $owner->Title;
+        $title = Convert::raw2xml($title);
+        $owner->extend('updateTreeTitle', $title);
+        return $title;
+    }
```
ChloeSartorelli, line 138: This fixes the specific unit test rather than the underlying regression.

If `updateTreeTitle` should permit HTML content, we should change where `raw2xml()` is called _in the method that calls it (in `Hierarchy`)_ so that other `updateTreeTitle` extension hooks can behave as expected

GuySartorelli (changes requested):

emteknetnz:
Closed in favour of https://github.com/silverstripe/silverstripe-framework/pull/11517

PR #624, opened 2025-01-07
API Remove deprecated API

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/311

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -313,13 +313,12 @@ class Subsite extends DataObject
      * @param class-string<T> $className
      * @param string $filter
      * @param string $sort
-     * @param string $join Deprecated, use leftJoin($table, $joinClause) instead
      * @param string $limit
      * @return DataList<T>
      */
-    public static function get_from_all_subsites($className, $filter = '', $sort = '', $join = '', $limit = '')
+    public static function get_from_all_subsites($className, $filter = '', $sort = '', $limit = '')
     {
-        $result = DataObject::get($className, $filter, $sort, $join, $limit);
+        $result = DataObject::get($className, $filter, $sort, $limit);
         $result = $result->setDataQueryParam('Subsite.filter', false);
         return $result;
     }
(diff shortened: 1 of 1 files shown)


src/Model/Subsite.php
```diff
@@ -932,13 +931,24 @@ public function getMembersByPermission($permissionCodes = ['ADMIN'])
...
+                . " IN ('$SQL_permissionCodes')"
+        )->leftJoin(
+            $groupMembersTable,
+            $schema->sqlColumnForField(Member::class, 'ID') . " = \"{$groupMembersTable}\".\"MemberID\""
+        )->leftJoin(
+            $groupTable,
+            $schema->sqlColumnForField(Group::class, 'ID') . " = \"{$groupMembersTable}\".\"GroupID\""
+        )->leftJoin(
+            $schema->tableName(Permission::class),
+            $schema->sqlColumnForField(Permission::class, 'ID') . ' = ' . $schema->sqlColumnForField(Group::class, 'ID')
```
GuySartorelli, line 951: Since I had to refactor this anyway, I decided to remove the hardcoded references since those weren't following best practices.
The query should ultimately be identical.

src/Model/Subsite.php
```diff
@@ -919,29 +918,6 @@ public function activate()
         Subsite::changeSubsite($this);
     }
 
-    /**
-     * @param array $permissionCodes
-     * @return DataList<Member>
-     */
-    public function getMembersByPermission($permissionCodes = ['ADMIN'])
```
ChloeSartorelli, line 927: This method was added in https://github.com/silverstripe/silverstripe-subsites/commit/5e9b67216f1191c0e19f5dda75ba6e685cd5335f which has no context whatsoever about what its intended use was or how it was intended to work.

Right now it just simply doesn't work. I had a choice between fixing it and removing it, and since there's no context about how it _should_ work and it's impossible for anyone to use right now anyway, fixing it just didn't seem worthwhile.

emteknetnz (approved):

PR #625, opened 2025-01-08
API Deprecate API which will be removed in CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/311

### DIFF
--- a/src/Model/Subsite.php
+++ b/src/Model/Subsite.php
@@ -312,7 +312,7 @@ class Subsite extends DataObject
      * @param class-string<T> $className
      * @param string $filter
      * @param string $sort
-     * @param string $join
+     * @param string $join Deprecated, use leftJoin($table, $joinClause) instead
      * @param string $limit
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
Subsites is loosing support in CMS 6, don't bother with this, unless there's a really strong reason for this, this should just be closed

GuySartorelli:
We said we'd make it CMS 6 compatible. The easiest way to do that is to just treat it as supported until the CMS 6 stable launch. 
It can't really be CMS 6 compatible if it's calling method with the arguments in the wrong places.

emteknetnz:
It'll still be CMS 6 compatible without any changes (unless I misunderstand this PR), though I do appreciate "just treat it the same" as a reason

Update: OK I see, the DataObject::get(..., $join) param will be removed in a later PR

emteknetnz (approved):

PR #641, opened 2025-04-23
ENH Cache result of whether subsites exist

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

The same query was being called ~63 times on a base install of kitchen sink when going to `/admin/pages`

**Do not merge - this should target the CMS 5.4 branch, right now it targets CMS 6**


emteknetnz:
Replaced by https://github.com/silverstripe/silverstripe-subsites/pull/643

PR #643, opened 2025-04-30
ENH Cache result of whether subsites exists

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

The following query is repeatedly called about ~10 times when going to `/admin/pages` on installer + subsites, and about ~60 times on a base install of sink

`SELECT CASE WHEN EXISTS(SELECT * FROM "Subsite" ) THEN 1 ELSE 0 END`

### DIFF
--- a/src/Extensions/SiteTreeSubsites.php
+++ b/src/Extensions/SiteTreeSubsites.php
@@ -53,6 +53,11 @@ class SiteTreeSubsites extends DataExtension
     private static $many_many_extraFields = [
         'CrossSubsiteLinkTracking' => ['FieldName' => 'Varchar']
     ];
+    
+    /**
+    * Used to cache the result of a heavily called database query
+    */
+    private bool $subsitesExist;
 
     public function isMainSite()
     {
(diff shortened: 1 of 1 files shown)


GuySartorelli:
@emteknetnz Is the ~10 and ~60 times before or after this PR? Can you please give me what to expect before _and_ after?

emteknetnz:
Both the ~10 and the ~60 are before this PR. The ~10 is silverstripe/installer + silverstripe/subsites, the ~60 is silverstripe/recipe-kitchen-sink

After this PR, from memory is was just 1 query as we are now using an in-memory cache

src/Extensions/SiteTreeSubsites.php
```diff
@@ -53,6 +53,11 @@ class SiteTreeSubsites extends DataExtension
     private static $many_many_extraFields = [
         'CrossSubsiteLinkTracking' => ['FieldName' => 'Varchar']
     ];
+    
+    /**
+    * Used to cache the result of a heavily called database query
+    */
+    private bool $subsitesExist;
```
ChloeSartorelli, line 60: Note that because extensions are used as singletons, and subsites existing is a global thing, it *is* appropriate to cache this as a member property as this means all pages this is being called on will get the benefit of the cache.

GuySartorelli (approved):
LGTM, I can't think of any edge cases this would break and seems like a sensible improvement.

PR #656, opened 2025-12-04
Update SitewideContentSubsites namespace

ishannz:
## Description

User Subsite namespace instead of SiteWideContentReport on SitewideContentSubsites 

Dev/build failed due to a duplicate file with the same namespace when using the SiteWideContentReport module.

## Issues

- #

## Pull request checklist

- [ ] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [ ] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [ ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/_config/extensions.yml
+++ b/_config/extensions.yml
@@ -75,4 +75,4 @@ Only:
 ---
 SilverStripe\Reports\SiteWideContentReport\Reports\SitewideContentReport:
   extensions:
-    - SilverStripe\SiteWideContentReport\Model\SitewideContentSubsites
+    - SilverStripe\Subsites\Model\SitewideContentSubsites
--- a/src/Extensions/SitewideContentSubsites.php
+++ b/src/Extensions/SitewideContentSubsites.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace SilverStripe\SiteWideContentReport\Model;
+namespace SilverStripe\Subsites\Model;
 
 use SilverStripe\Core\Extension;
 use SilverStripe\SiteWideContentReport\Form\GridFieldBasicContentReport;


GuySartorelli:
`silverstripe/sitewidecontent-report` isn't compatible with this version of subsites (i.e. CMS 6), so that shouldn't be a problem.

ishannz:
> `silverstripe/sitewidecontent-report` isn't compatible with this version of subsites (i.e. CMS 6), so that shouldn't be a problem.

Could you please explain why it uses a different namespace (SiteWideContentReport) in the subsite module?

GuySartorelli:
We copied the report into here since `silverstripe/sitewidecontent-report` isn't going to be made CMS 6 compatible.

The namespace should have changed at that time but we missed it. But it doesn't affect anything, since you can't install `silverstripe/sitewidecontent-report` in CMS 6 anyway.

Changing the namespace would be a breaking API change so I would recommend this change not be merged.
That said, since this module is no longer commercially supported, the [release policy](https://docs.silverstripe.org/en/6/project_governance/release_policy/) and [definition of public API](https://docs.silverstripe.org/en/6/project_governance/public_api/) don't apply, so if someone with merge access wants to merge this they can.
Though it looks like there are no community maintainers for this module currently.

